PR #73, opened 2014-06-18
Add matplotlib 1.3.1

iamaziz:
Here you go.
And the docset is here:
http://goo.gl/5edDnH 

Thank you :+1:


tlkahn:
It is gone, again. Should we re-open it?

Kapeli:
Matplotlib is now under Preferences > Downloads > Main Docsets.

AllanLRH:
I can't see it in the preferences (still on Dash 2, but considering upgrade).
Is it because I use Dash 2? And can I fix it by adding the docset manually somehow?

PR #282, opened 2015-01-16
Add Chrome Extensions API

gebrkn:


Kapeli:
Looks good.

Please change the name of the docset inside Info.plist to "Chrome Extensions API" and also the bundle and platform identifiers to just "chrome".

You should use Chrome_Extensions_API only for names that are not displayed to the user directly (i.e. the docset archive, the docset project folder and so on).

gebrkn:
I'm a bit unsure, is this correct?

```
	<key>CFBundleIdentifier</key>
	<string>chrome</string>
	<key>CFBundleName</key>
	<string>Chrome Extensions API</string>
	<key>DashDocSetFamily</key>
	<string>javascript</string>
	<string>chrome</string>
	<key>dashIndexFilePath</key>
	<string>index.html</string>
	<key>isDashDocset</key>
	<true/>
```

If yes, I'll submit a new PR

Kapeli:
Yep, that's fine.

PR #298, opened 2015-01-24
Update Hammerspoon docset to 0.9.19

cmsj:


cmsj:
(btw, I hope I'm not overwhelming you with these updates. If there's a better way I should be getting docs into Dash, I'd be happy to learn about it)

Kapeli:
Don't worry about it. All I need to do is press the Merge button and then run a command on my Mac.

Post as many updates as you want/need.

PR #301, opened 2015-01-25
Haiku docset update 0.4.8

return:
API changes to:
BInvoker, BMessage, BMessageRunner, BJoyStick, BSize, BToolTip,
BTab,BTabView, BOptionPopUp and BTextControl


Kapeli:
Thank you! :+1:

PR #303, opened 2015-01-26
Add Pandoc - document converter

iamaziz:


Kapeli:
What's up with the `0.0` version? If you don't want a docset to display a version at all, start the version with a `/`, like this: `/blabla`.

iamaziz:
gosh! forget that while I was looking up the version info. It is actually 1.13.2 http://pandoc.org/releases.html , should I do the pull again?

iamaziz:
"deep learning" should not be in the aliases too! I will redo the pull

PR #305, opened 2015-01-27
Improve Theano docset

royxue:
Hi,

I'm a contributor of Theano, I made this PR to improve the existing docset.

I have asked Aziz, he generate the doc from our website, however we prefer to use the doc generated from our dev repo, we think this version would be more useful. It's still 0.6 version, but somewhere ahead of the website version.

PS: do I need to delete the .tgz.txt file?


Kapeli:
There's no need to delete the .tgz.txt file.

Thank you! :+1:

PR #309, opened 2015-01-28
Contributing docset for Eigen 3.2.4

elrnv:
Eigen is a C++ linear algebra library: http://eigen.tuxfamily.org/


Kapeli:
Please don't include the version number in the docset name (in the docset's Info.plist). Just set that to "Eigen".

elrnv:
I couldn't find the difference between the docset in bef5e65 and a9973a6. All I did was recompile the docset.
On another note, some of the doxygen generated pages look different than in the [official documentation](http://eigen.tuxfamily.org/dox/). For instance the width of the [Matrix reference page](http://eigen.tuxfamily.org/dox/classEigen_1_1Matrix.html) extends farther in the generated docs. For some reason the Public Member Functions table doesn't wrap the return types.
I'm not sure how to fix this problem at this time.

Kapeli:
Thank you! :+1: 

> On another note, some of the doxygen generated pages look different than in the official documentation

Must be a CSS difference as far as I can tell. Unfortunately I don't have time to look into it right now :(

PR #310, opened 2015-01-30
Initial commit of Docker docset

MWers:


Kapeli:
Something changed in the YouTube embed code and it now is:

```html
<iframe width="640" height="360" src="//www.youtube.com/embed/wQsrKX4588U?rel=0" frameborder="0" allowfullscreen></iframe>
```

Instead of
```html
<iframe width="640" height="360" src="http://www.youtube.com/embed/wQsrKX4588U?rel=0" frameborder="0" allowfullscreen></iframe>
```

As it's a local file, you need to include the `http` bit as well, otherwise Dash tries to look for youtube.com locally.

I'm going to merge this since in the current version of Dash YouTube videos don't work anyway, but please update it sometime to get the docset ready for the next Dash update which will support YouTube videos.

MWers:
Good catch. The `<iframe>` code in the Docker docs uses [protocol-relative urls](http://www.paulirish.com/2010/the-protocol-relative-url/) although if the containing page is requested using the `file://` scheme (as I'm guessing the Dash docs are), then it would try to request the video as a local file and fail. I'll make the change today.

PR #314, opened 2015-02-03
pin docset

TheCjw:
I've created a docset of [pintool](https://software.intel.com/en-us/articles/pin-a-dynamic-binary-instrumentation-tool), including xed32 and xed64. Tested on Dash for Mac v2.2.3.


Kapeli:
Failed validation: a lot of your entries don't have a "type" set.

PR #318, opened 2015-02-04
Repull pin docset

TheCjw:
Fix the issue that there're some entries didn't has the `type` set, 148 entries were updated to `define.`


Kapeli:
Thank you! :+1:

PR #337, opened 2015-02-14
Sublime Text Plugin API docset

leonid-shevtsov:
This is a docset for Sublime Text plugin developers. It contains an indexed reference of available classes and modules, based on the official API, but enhanced with version compatibility information.


Kapeli:
Thank you! :+1:

PR #339, opened 2015-02-17
Docset for Susy CSS grid framework

bookwyrm:


Kapeli:
You need to add a copy of `Susy.tgz` inside the main project folder as well. Dash doesn't use the versioned docset automatically, so it needs to be duplicated at `docsets/Susy/Suzy.tgz` for things to work.

bookwyrm:
Thanks. I've added 'Susy.tgz` to the main project folder as well.

Kapeli:
Thank you! :+1:

PR #349, opened 2015-02-25
Add Shogun Machine Learning Toolbox docset

vigsterkr:


Kapeli:
Why not call it just Shogun in the `docset.json` name?

vigsterkr:
we could, but this is the full name of the library :)

Kapeli:
Fair enough :+1:. Thanks for the contribution!

PR #350, opened 2015-02-25
Update Three.js r70 docset

toruta39:


Kapeli:
Please redo your fork and squash your commits into one.

PR #351, opened 2015-02-25
add octave docset in version 3.8.0

obstschale:
Hi,

some time ago I created this docset and I also tried to submit it once (#13). Unfortunately, I forgot it. But here is my second attempt :)


Kapeli:
Thank you! :+1:

PR #358, opened 2015-02-28
Added Wand docset

joaoponceleao:
Docset for Wand: https://github.com/dahlia/wand
The currently maintained ImageMagick binding for Python.
May add a script for next wand version if needed.


Kapeli:
Your `docset.json` points to `wand.tgz`, but the file is actually `Wand.tgz`. Please fix it.

joaoponceleao:
Will do. Also have to change the resolution of the icon inside the docset. Question though:
The aliases list in docset.json. Is it purely for aliases of the actual docset? Or can it be used as follows (considering this is an ImageMagick binding for Python):
'MagickWand' (binding last updated in 2008)
'PythoMagick' (alternative binding)
'ImageMagick' (when user searches for ImageMagick, this would also return all the available bindings / interfaces for the library).

Kapeli:
You can include any alias that you think is relevant. All of the aliases you asked about are fine. A `Python` alias would be ok too.

One more thing: the name inside the `docset.json` should be changed from `Wand Docset` to just `WanPlease note that you can generate your own docsets - http://kapeli.com/docsets..

joaoponceleao:
Ok. Think everything should be fine now.

Kapeli:
Thank you! :+1:

PR #361, opened 2015-03-03
Submit again

TongG:
How about now?


Kapeli:
All good. However, are you aware you could make this a Cocoa Pod and then it would show up on CocoaDocs.org? Then it would appear in Dash in Preferences > Downloads > Cocoa Docsets too.

TongG:
Thank you for your tip, I will make this framework Cocoa Pod when the next version be released. That's OK this time. 😃

Kapeli:
If you'll make it a Cocoa Pod, then there's no point in accepting this as I don't want the same docset to appear in two places. Sorry.

TongG:
Fine, I'll resubmit pull request after completing it. Thank you.

Kapeli:
You don't have to resubmit anything here. Just add Cocoa Pod support and it should automatically show up on Cocoa Docs.

TongG:
Yep!

PR #364, opened 2015-03-05
Serverspec docset

coderanger:
This adds a docset-ized version of serverspec.org. Unfortunately because the docs live in their own repo, there don't seem to be any tags or other way to come up with a version number so I've set it to `/1`.


Kapeli:
Thank you! :+1:

PR #376, opened 2015-03-16
Haiku docset update

return:
Changes to ScrollView and BReferencable classes


Kapeli:
You need to update the docset.json as well, otherwise you're not really changing anything.

return:
Thanks, it is now updated.

Kapeli:
Thank you! :+1:

PR #382, opened 2015-03-20
Hapi 8.4.0

z0mt3c:


Kapeli:
Thank you! :+1: 

Next time, please stop versioning your docset archives. There's no need to. Just rename to something like `hapi.tgz`. Thank you!

z0mt3c:
kk ;)

PR #389, opened 2015-03-25
added Apache jclouds document

itachi90:
Added Apache jclouds document, which is useful for Java developer focusing on cloud platform development especially with OpenStack.


Kapeli:
This seems to be the same with the docset in Preferences > Downloads > Java Docsets > jclouds. Sorry.

PR #394, opened 2015-04-01
update Ramda docset

alexbepple:
Not sure what the proper procedure is for updates. Hopefully this works. If not, please let me know.

Particularly, I was not sure whether to leave the .tgz.txt.


Kapeli:
The .tgz.txt doesn't matter. You can remove it or leave it alone, up to you.

PR #398, opened 2015-04-03
New Vaadin 7.4.3

ewall:
Well, I guess I can't complain that the Vaadin folks don't update often enough! Here's another minor version update.

Please note that I have been getting some errors from javadocset when generating these; I just [submitted a log as an issue here](https://github.com/Kapeli/javadocset/issues/6).

And, as always, thanks for making such a useful tool!


Kapeli:
Please redo your fork.

PR #400, opened 2015-04-04
Fixed versions of the 0.7.5 archive.

dazoakley:
These now contain the CSS/JS and image assets for running offline.


Kapeli:
Thank you! :+1:

PR #401, opened 2015-04-04
Fixed the archive, and updated to 0.4.0.

dazoakley:
These now contain the CSS/JS and image assets for running offline.


Kapeli:
Please also update the `docset.json` with the new version.

dazoakley:
All updated, forgot about the `docset.json`. :)

Kapeli:
Thank you! :+1:

PR #406, opened 2015-04-06
add LiveScript docset

alexbepple:


alexbepple:
I checked the build log.

> Found regular icon, but missing retina icon in "LiveScript"

I do not have access to a retina icon right now. Should I leave things as they are or upscale the regular icon?

> Could not load entries at "LiveScript.docset"

Not sure what this means. I checked the contents manually. They seem fine to me.

Kapeli:
> I do not have access to a retina icon right now. Should I leave things as they are or upscale the regular icon?

You should remove the icon completely. You must have both a regular and a retina icon, or none at all.

> Could not load entries at "LiveScript.docset"

This means your docSet.dsidx is empty (contains no entries). This can also be seen when you try to expand your docset in Dash's docset browser and Dash says "Docset is empty".

alexbepple:
Alright. I thought I would make the first public release without an index.

You should make clearer that an index is required. [The checklist](https://github.com/Kapeli/Dash-User-Contributions/wiki/Docset-Contribution-Checklist) only says that the index must not contain invalid entries.

Issue #407, opened 2015-04-07
Atom docset outdated


bkbooth:
I'm the contributor of the Atom.io docset, it's quite outdated now and I don't have the time to keep up with the changes.

Perhaps it might be best to remove this docset for now and I'll take another look at it once they reach a 1.0 version and hopefully the changes to the API slow down a little.

Also happy for anyone else to take over this if they have time/interest.

Kapeli:
I'll leave it alone for now. I might remove it if it bothers anyone in the future. Thanks for the heads up!

Kapeli:
I'll leave this issue open in case anyone wants to take over.

PR #408, opened 2015-04-08
Update Atom 0.190.0

Ephemera:
I'm interested in taking over #407.


Kapeli:
Good job.

Some issues that need to be fixed:

1. Your `dashIndexFilePath` entry points to `index.html` (in Info.plist), which redirects to `Atom.html`. Please change `dashIndexFilePath` so that it points to `Atom.html` directly
2. The doc pages load a CSS file and a JS file from the web. If possible, please change this so that all resources are loaded locally (and available while offline).
3. As far as I can tell, none of the Method entries are linked correctly (they point to anchors that don't exist in the pages). For example, `Atom::beep` points to `Atom.html#beep`, while it should point to `Atom.html#instance-beep` instead
4. The order of the `specific_versions` is wrong. The latest version must be at the top, while the oldest version should be at the bottom. `0.190.0` should be listed before `0.120.0`

Thank you!

PR #409, opened 2015-04-09
Update Atom 0.190.0

Ephemera:
with fixing [it][1]

[1]: https://github.com/Kapeli/Dash-User-Contributions/pull/408#issuecomment-91020620


Kapeli:
The `dashIndexFilePath` you've set is not valid anymore after your changes to support offline CSS. The `dashIndexFilePath` should be `docs/api/v0.190.0/Atom.html`. Please fix this and also please remember to change this with each docset update.

PR #412, opened 2015-04-10
Update React Native docset for version 0.3.4 and minor changes per the guideline 

epitaphmike:


Kapeli:
Great. Thank you!

One suggestion: the docset seems to load something online. If possible, all resources should be loaded locally.

epitaphmike:
There are links to a few external resources. I'll go through and clean those up for the next release.

PR #415, opened 2015-04-10
Add D language docset

trishume:
This adds a docset for the [D language](http://dlang.org/) standard library (sometimes referred to as Phobos).

It is generated with [my generator script](https://github.com/trishume/d-phobos-docset) which downloads and scrapes the [online documentation](http://dlang.org/phobos). This script creates index entries for all the different pages (individual modules) and entries on the pages. It also tweaks the stylesheet to always activate the mobile mode designed for narrow screens for easier viewing in Dash. It has an index page and should work entirely offline.


Kapeli:
Good job! There seems to be an issue with Function entries as they point to non-existent anchors inside the pages. For example, all Function entries with the name "add".

PR #416, opened 2015-04-10
Add D language docset

trishume:
This is the same as my old PR except run with a new version of my generator script that fixes the bug you found in #415. My script didn't properly handle class members and tried to link to an anchor including only the field and not the class name as well. I have updated my script to scrape for anchors so now all anchors should exist and methods should be properly associated with their class.

### Old PR
This adds a docset for the [D language](http://dlang.org/) standard library (sometimes referred to as Phobos).

It is generated with [my generator script](https://github.com/trishume/d-phobos-docset) which downloads and scrapes the [online documentation](http://dlang.org/phobos). This script creates index entries for all the different pages (individual modules) and entries on the pages. It also tweaks the stylesheet to always activate the mobile mode designed for narrow screens for easier viewing in Dash. It has an index page and should work entirely offline.


Kapeli:
Thank you! :+1:

PR #420, opened 2015-04-17
add LuaNginxModule docset

ruoshan:
the original doc is here: https://github.com/openresty/lua-nginx-module/blob/master/README.markdown


Kapeli:
Thank you! :+1:

PR #427, opened 2015-04-22
Added v1.6.0 of Docker docs

MWers:
Added v1.6.0 of Docker docs, including section TOCs. 

Question: How would I get these docs added to the list of docs available to the Dash iOS app?

Thanks!

### DIFF
--- a/docsets/Docker/docset.json
+++ b/docsets/Docker/docset.json
@@ -1,6 +1,6 @@
 {
     "name": "Docker",
-    "version": "1.5.0",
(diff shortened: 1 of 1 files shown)


Kapeli:
Thank you! :+1: 

> Question: How would I get these docs added to the list of docs available to the Dash iOS app?

You can't. The iOS version doesn't support user contributed docsets right now. It's high priority on my todo list though, so I should get to it as soon as I'm done with the next OS X update.

PR #430, opened 2015-04-28
update RFCs docset to v1.2

willnorris:
This docset is over the 100 mb limit, so I've not included the tarball in this commit. The current 1.2 version can be obtained from https://github.com/willnorris/rfcdash/releases/download/v1.2/RFCs.tgz

fixes willnorris/rfcdash#3


Kapeli:
Thank you! :+1:

PR #432, opened 2015-04-29
Update OpenUI5 Docset

monavari-lebrecht:


Kapeli:
In your `docset.json` the latest versions in `specific_versions` should be at the top (i.e. `1.28.5` should appear before `1.20.8`). I'll fix it this time.

PR #435, opened 2015-05-04
Updated Phalcon + Added Phalcon 2 docset

saimusdev:


Kapeli:
Please redo your fork and your pull request.

Also please don't submit Phalcon 2 as a separate docset. Just post is as an update to the existing Phalcon docset.

If you still want users to be able to install Phalcon 1.x, then you should add Phalcon 1.x under `specific_versions` in your `docset.json`.

PR #437, opened 2015-05-04
Updated Phalcon docset (1.3.0 ---> 2.0.0)

saimusdev:


Kapeli:
That's not how you define `specific_versions`. Follow the sample at https://github.com/Kapeli/Dash-User-Contributions/blob/master/Sample_Docset/docset.json.

PR #438, opened 2015-05-05
Posting docset

ColdfusionSun:
Revised and corrected docset.json revised first submit of CFWheels


Kapeli:
The docset is not archived correctly. The archive should **only** contain the docset and nothing else. Your archive contains a directory structure (i.e. `Users/james/Desktop/CFWheels.docset`).

Use this command to archive:
```
tar --exclude='.DS_Store' -cvzf <docset name>.tgz <docset name>.docset
```

PR #439, opened 2015-05-05
Updated Phalcon docset for version 2.0.0

saimusdev:


Kapeli:
Thank you! :+1: 

Note for next time: it's not enough to include the docset archive for the `specific_versions`, you have to include it for the main thing as well (i.e. `docsets/Phalcon/versions/2.0.0/Phalcon.tgz ` should also be at `docsets/Phalcon/Phalcon.tgz`).

saimusdev:
Noted

PR #443, opened 2015-05-07
Kafka 0.8.2 docset

Igosuki:
Was able to untar a newly made archive with the same command as dash', don't know what went wrong previously.


Kapeli:
The issues are:

```
Found entry with new line in its name at row id "40" at "kafka.docset". Complete entry name is: "Mirroring data between
                    clusters"
Found entry with new line in its name at row id "41" at "kafka.docset". Complete entry name is: "Expanding your
                    cluster"
Found entry with new line in its name at row id "42" at "kafka.docset". Complete entry name is: "Decommissioning
                    brokers"
Found entry with new line in its name at row id "43" at "kafka.docset". Complete entry name is: "Increasing
                    replication factor"
```

Unarchival issue: no idea. Can you archive using the same way as you archived Kafka/versions/0.8.2/kafka.tgz? That one seems to work.

Igosuki:
The archive at the root of the docset ? Sure

Kapeli:
Yes, the archive at the root doesn't work for some reason. The archive at `Kafka/versions/0.8.2/kafka.tgz` works.

Igosuki:
The blob doesn't leave any differences, I'll just overwrite it with two commits to make sure.

PR #445, opened 2015-05-09
Added GNU Make 4.1 docset

benzado:
Surprised this wasn't already here!


Kapeli:
You've put the docset archive at the wrong path. It should be at `docsets/GNU_Make/GNU_Make.tgz` instead of `docsets/GNU_Make/versions/4.1/GNU_Make.tgz`.

PR #447, opened 2015-05-10
Kafka docset

Igosuki:
Regenerate kafka archive with proper tar -cvzf

I regenerated both archives using the same command.


Kapeli:
The archives are fine, but you didn't fix the entries that contain new lines:

```
Found entry with new line in its name at row id "40" at "kafka.docset". Complete entry name is: "Mirroring data between
                    clusters"
Found entry with new line in its name at row id "41" at "kafka.docset". Complete entry name is: "Expanding your
                    cluster"
Found entry with new line in its name at row id "42" at "kafka.docset". Complete entry name is: "Decommissioning
                    brokers"
Found entry with new line in its name at row id "43" at "kafka.docset". Complete entry name is: "Increasing
```

Igosuki:
Right. I'll just do the whole build of Dash on my side end 2 end before ever submitting again, I feel like a fool doing 1 PR a day ;)

PR #456, opened 2015-05-15
Added squash requirements to the README.md

WoLpH:


Kapeli:
Squashing isn't a requirement. It's only a requirement when you move around large files (i.e. the archived docsets), because it makes cleaning up a lot harder.

PR #460, opened 2015-05-18
Integrate DokuWiki documentation

gamma:
This is not an official version, but uses the same documentation as http://xref.dokuwiki.org/reference/dokuwiki/


Kapeli:
Thank you! :+1:

PR #462, opened 2015-05-20
added CouchDB 1.4 docset

SteveBenner:


SteveBenner:
[The build appears to have failed](https://travis-ci.org/Kapeli/Dash-User-Contributions/builds/63413514)

Kapeli:
Issues are:
```
Empty path found for row id "59" at "CouchDB-1.4.docset"
Empty path found for row id "60" at "CouchDB-1.4.docset"
Empty path found for row id "61" at "CouchDB-1.4.docset"
Empty path found for row id "62" at "CouchDB-1.4.docset"
Empty path found for row id "63" at "CouchDB-1.4.docset"
Empty path found for row id "64" at "CouchDB-1.4.docset"
```

This means that your docSet.dsidx contains entries with null paths.

SteveBenner:
I’m not sure how I would begin to troubleshoot this issue; can you give me a pointer?

Kapeli:
Your generation script fails somewhere and inserts entries that do not have a path. See screenshot below.

screen shot 2015-05-21 at 03 09 04

Also, please note that you should NOT include the unarchived docset anywhere. Just include the `.tgz`.

SteveBenner:
I see, thank you. I thought that I had tested the generated docset thoroughly when I first uploaded it, but apparently I somehow missed those nulls. Revising the script… 

Also, inclusion of the unarchived docset was accidental; it will of course be removed for my next pull request.

Kapeli:
Also one more issue: your docset.json points to `CouchDB.tgz` which doesn't exist. You have `CouchDB-1.4.tgz` instead.

SteveBenner:
Yes, I caught the error in `docset.json`, and it brings up a point I wanted to ask you about, actually... I was wondering what you thought about including the version name in the docset title in cases like CouchDB’s. The reason I initially did so, and ultimately decided this was correct, is because CouchDB’s docs vary widely between minor versions, both in content and arrangement. I’ve no idea whether CouchDB adheres to semantic versioning or not, but I know that most, if not all of the minor releases since 1.4 would require their own unique docsets (something on my todo list now).

The problem with my generation script turned out to be an edge case in parsing one of the entry types… Easy to fix, but also easy to miss. Your very helpful comment made me realize that reviewing the contents of aforementioned `docSet.dsidx` file is an great way to verify the outputted docset data. It might be a good idea to put a mention of this on your website page for developing docsets. I know it would have helped me avoid this mistake in the first place, as I read everything available on your site. But then again, I don’t deal with SQL very much, so this might be a totally obvious thing… Just a thought!

P.S. I found the app [SQLPro for SQLite Read-Only][1] to be a great tool for viewing the `.dsidx` file. I mention this because **it’s free** :smiley: 

[1]: https://itunes.apple.com/us/app/sqlpro-for-sqlite-read-only/id635299994?mt=12

Kapeli:
> I was wondering what you thought about including the version name in the docset title in cases like CouchDB’s

You shouldn't. In Dash, docsets that don't show a version are docsets that get updated and the goal is to assume that they're always the latest version. Docsets that show a version are docsets that are locked at that version and don't get updated.

You shouldn't include the version even in the `specific_versions` docsets, because Dash will add it itself while installing.

> Your very helpful comment made me realize that reviewing the contents of aforementioned docSet.dsidx file is an great way to verify the outputted docset data. It might be a good idea to put a mention of this on your website page for developing docsets.

I like to keep things simple and not confuse users. These issues happen rarely, so I don't really see a need to clutter the instructions right now. I might change my mind in the future though.

SteveBenner:
Unsure of whether you grasped the nature of my concern with distributing multiple versions of CouchDB docsets, which is that even between **minor** versions (e.g. 1.4 and 1.5) the contents differ enough to necessitate unique docsets for each, I looked to existing uploads for the answer. I discovered that versioned docsets are in fact represented separately in the interface, so my question was taken care of.

**I would like your help** in confirming the following JSON as valid for use in `docset.json` file. Much thanks!

```json
{
    "name": "CouchDB",
    "version": "1.4",
    "archive": "CouchDB.tgz",
    "author": {
        "name": "Steve Benner",
        "link": "http://sensiblesolutions.me"
    },
    "aliases": [],
    "specific_versions": [
        {
            "_comment": ""
        },
        {
            "version": "1.4.0",
            "archive": "versions/1.4/CouchDB.tgz"
        }
    ]
}
```

I also have another question, though it’s just a formality: would you recommend using `Apache CouchDB` in place of `CouchDB` since that is technically the full name of the software? If so, where exactly?

Kapeli:
> Unsure of whether you grasped the nature of my concern with distributing multiple versions of CouchDB docsets, which is that even between minor versions (e.g. 1.4 and 1.5) the contents differ enough to necessitate unique docsets for each, I looked to existing uploads for the answer 

I understand your concern, but this is how Dash handles displaying versions for all docsets and it has never been an issue for users. If users start complaining about the version not being shown for CouchDB, we can fix things later.

> I would like your help in confirming the following JSON as valid for use in docset.json file. Much thanks!

The JSON looks fine. Please note that even `versions/1.4/CouchDB.tgz` should NOT contain a version in the Info.plist. Dash will add that version when the docset is installed.

> I also have another question, though it’s just a formality: would you recommend using Apache CouchDB in place of CouchDB since that is technically the full name of the software? If so, where exactly?

Using the full name is not required, just use the most common one. CouchDB should be fine in my opinion.

However, you should add "Apache CouchDB" as an alias in `docset.json`. Doing this will ensure that if a user searches for "Apache CouchDB" in Preferences > Downloads, your docset will still be shown.

```json
"aliases": ["Apache CouchDB"],
```

SteveBenner:
> I understand your concern, but this is how Dash handles displaying versions for all docsets and it has never been an issue for users.

To clarify, I wanted to know if Dash would display multiple versions of a docset separately, which is something that CouchDB would require. I realized I could find the answer myself, by downloading multiple versions of one of the existing user-contributed docsets. The current behavior is just what I had in mind; I don’t think anything needs to be changed.

Kapeli:
Ah, I see. Let me know if you notice any issues in the future or have any suggestions :+1:

PR #464, opened 2015-05-21
Add docset for GNU Guile 2.0

pekingduck:
Hi,
I created a "GNU Guile Reference Manual 2.0". 

Thanks.


Kapeli:
Your `docset.json` points to `guile.tgz`, but that file doesn't exist. I think you meant `GNU_Guile.tgz` instead.

PR #471, opened 2015-05-27
ionic 1.0.0/final

DarkSwoop:
ionic 1.0.0 has been released. Here is the updated docset.


Kapeli:
Failed validation:

```
File not found at "docs/nightly/api/directive/ionSlidePager/index.html" for entry with row id "85" of "Ionic.docset"
```

This means that the entry in docSet.dsidx at row 85 points to a file that does not exist.

PR #478, opened 2015-05-29
Added docset for CodeIgniter 3.0.0

schwemmer:


Kapeli:
A CodeIgniter docset made by me is already available in Preferences > Downloads. Sorry.

schwemmer:
Argh..
Image of Fry saying Argh

You're right. I didn't see the little arrow on the right of the download button that let's you choose the version. My bad. I guess at least I learned how to make a docset. 

Thanks anyways for making an awesome tool that helps me everyday!

Kapeli:
You don't really need to use the little arrow to get CodeIgniter 3.0.0 though. If you just press Download, it should install 3.0.0 (the latest version).

schwemmer:
The thing is, I already had a 2.x docset installed and then when I searched for CodeIgniter 3 it looked like the one I already had installed. I should have noticed the arrow...

Is the [Swift Language Guide](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/TheBasics.html#//apple_ref/doc/uid/TP40014097-CH5-ID309) available as a docset? I would find it useful and I can't seem to find it as a docset. If not, I might try to create one, now that I know how :-)

Kapeli:
The Swift Language Guide is part of Apple's Xcode docset, but it's not indexed apart from the entry for the whole guide... You could definitely make a docset for it if you want :)

schwemmer:
I gave it a try but can't get it to index properly for some reasons.

PR #489, opened 2015-06-10
update ClojureScript

shaunlebron:
- add syntax/destructure entry
- correct type member history
- correct ExceptionInfo type
- start ignoring unlabeled internals
- stylize type members in ref table


Kapeli:
Thank you! :+1:

PR #490, opened 2015-06-10
Added pyspark docset

ronert:


Kapeli:
Failed validation:

1. You didn't archive the docset correctly. You need to archive the docset directly. Instead, you archived a folder that contains the docset.
2. You need to provide a `icon@2x.png`. The one you included seems to be the one in the sample project.
3. Delete `versions/1.0.0/Sample_Docset.tgz` and `versions/1.0.1/Sample_Docset.tgz`.

PR #493, opened 2015-06-12
Update Drush Docset

mikebell:
Updated the drush docset with new commands.


Kapeli:
Please redo your fork and squash all the commits into 1.

PR #495, opened 2015-06-12
Add docset for Kirby cheat sheet

samnabi:
Kirby cheat sheet version 2.1.0


Kapeli:
Failed validation. The paths still don't work.

Please note that you're not supposed to include the unarchived docset, only the archived one.

Kapeli:
Also, as this is a cheat sheet, you might want to consider converting it to a Dash cheat sheet instead: https://github.com/Kapeli/cheatsheets#readme.

samnabi:
Ah, I updated the paths but forgot to re-create the archives.

There are other sections of the Kirby docs that don't fit well into the cheatset format, so I think I'll stick with this format.

PR #500, opened 2015-06-14
update ClojureScript

shaunlebron:
- separate syntax API section
- add special symbols (Infinity, NaN, true, false)
- add special namespaces (js/, Math/)
- add destructure forms (:keys, :strs, :syms, :or, :as, &)
- add sub-syntax forms (#?@, ~@, ::)
- mark syntax forms that are edn-compatible


Kapeli:
Thank you! :+1:

PR #501, opened 2015-06-14
Fixed pyspark docset

ronert:


Kapeli:
Same issue as before, it's not archive correctly. Try using this command:

```
tar --exclude='.DS_Store' -cvzf <docset name>.tgz <docset name>.docset
```

Also, please redo your fork and squash all commits into one.

ronert:
The command I used is `tar --exclude='.DS_Store' -cvzf pyspark.tgz pyspark/pyspark.docset/` from the manual. How can I fix it?

Kapeli:
`cd` into pyspark and then `tar --exclude='.DS_Store' -cvzf pyspark.tgz pyspark.docset`

PR #506, opened 2015-06-15
Fixed a trucation error and bumbped version to 1.4 so it is in sync with Spark versioning

ronert:


Kapeli:
Please redo your fork and squash all commits into one.

PR #516, opened 2015-06-26
Add DPDK

qhsong:
DPDK is a set of libraries and drivers for fast packet processing. I often use it so I merge it into Dash Program.I also write a go programmer to generator searchIndex.I think I can always maintain this document. Also,this is my first pull a request,I hope it can merge into you source code. Thx.


Kapeli:
Failed validation:
```
Empty path found for row id "947" at "DPDK.docset"
```

This means that the entry with id "947" in docSet.dsidx has an empty path.

Also what's up with `docsets/versions/1.0.0/Sample_Docset.tgz`? Why did you include them?

qhsong:
Ops,My script include some empty path.I will fixed it.
when I copy Sample_Docset,some wrong happens to me.I will fix it .Sorry for that.
thx

PR #529, opened 2015-07-05
Added new docset for Typo3 CMS

cgrossde:
I created docsets for all currently used Typo3 versions plus the master. Could you please add them?
The script that was used to generate these docsets can be found here: https://github.com/cgrossde/Typo3Docsets

PS: Is it a problem that one version is named `master`?

PPS: According to [this issue](https://forge.typo3.org/issues/37427) there was once a Typo3 docset. Do you know what happened to it?


Kapeli:
TYPO3 is one of the officially supported Dash docsets. It can be installed from Preferences > Downloads > All Docsets > TYPO3. The only version available currently is 6.2. If you require any other versions let me know and I'll add them. Please note that I only support stable versions.

There's no point in having TYPO3 in the user contributed docsets.

If you'd like to contribute something like TYPO3 Flow (I noticed you mentioned it in the issue), please open a separate pull request.

cgrossde:
Holy s***, I could have sworn I searched it several times and did not find it. However it is quite useful to have multiple versions of Typo3 available since 4.5 LTS and all the other releases till 6.2 are still heavily used. Typo3 7.x is currently under heavy development, so a master branch would be still useful. Maybe you can think about adding more versions of Typo3.

I will make a pull request for Flow once it's done building (creating `.docsets` is quite slow ...)

Kapeli:
Great. I've added a todo entry to add more versions for TYPO3. Should get to it soon-ish.

Kapeli:
Docsets for TYPO3 6.2, 6.1, 6.0, 4.7, 4.6 and 4.5 are now available in Preferences > Downloads.

cgrossde:
You are awesome! Thanks.

PR #530, opened 2015-07-05
Added Typo3 FLOW docsets

cgrossde:
New docset for Typo3 FLOW :)


Kapeli:
Few issues:

1. Rename everything to `TYPO3 Flow` (for things that are displayed to the user, including the docset name in `Info.plist`) or `TYPO3_Flow` (for file/folder names)
2. The `DocSetPlatformFamily` should **not** include a version number
3. You need to include an archive in your project's root folder as well, for the main docset. Considering the naming that is used now, that would be including an archive at `docsets/Typo3FLOW/Typo3FLOW.tgz`.
4. Backslashes in version numbers are only used to trigger updates without changing the displayed version number, so it doesn't make sense to use them here. You should remove all backslashes in all the version numbers.

When submitting your pull request again, please make sure to squash all commits into a single commit.

PR #534, opened 2015-07-08
New docset for http://neo4j.com/docs/stable/cypher-refcard/

wires:
Hi,

This is  a pullrequest for a new docset containing the reference sheet for the Neo4J graph database query language called Cypher.

It only includes the stable version and since the HTML documentation doesn't contain any anchors, I've only added one entry in the index.

Please look critically at this as it is my first docset contribution :baby:

Thanks,
Jelle.


Kapeli:
A docset with only one entry isn't really useful... You need to add the anchors using your indexing script if the doc page doesn't have anchors. Alternatively, contact the maintainers of the docs and ask them to add anchors.

Other thoughts:
1. Everything needs to be renamed to just "Cypher". Most docsets are about References, so there's no need to have the name as "Cypher Reference" instead of just "Cypher"
2. You included the icons from the sample docset. Either add proper icons or delete the sample icons so that no icon is used
3. The doc page you are indexing looks a lot like a cheat sheet. Maybe it would be worth converting the page to Dash's cheat sheet format? See https://github.com/Kapeli/cheatsheets#readme.

PR #539, opened 2015-07-19
Adobe Experience Manager docset

nachocual:
I created these docsets for the most widely used versions of Adobe Experience Manager (AEM, a.k.a. CQ or CQ5). A lot of AEM devs use Macs, so I though it might be worth contributing them. Let me know your thoughts.


Kapeli:
Thank you! :+1:

PR #546, opened 2015-07-21
Adding Jenkins Job Builder docset

nextrevision:


Kapeli:
Thank you! :+1: 

One note: it would be better if the name inside Info.plist would be `Jenkins Job Builder` instead of `Jenkins_Job_Builder`, as this name is displayed to users. Not a huge issue though.

PR #547, opened 2015-07-22
IntelASMx86 docset

JeremyAgost:
Used HTML files from http://www.felixcloutier.com/x86/ to generate a dash docset package


Kapeli:
Thank you! :+1:

PR #549, opened 2015-07-22
Updateing JJB docset to include plugins and fixing displayed name

nextrevision:
Adding module plugins as "Plugin" type to search index. This also fixes the naming from Jenkins_Job_Builder to "Jenkins Job Builder".


Kapeli:
Why did you use `/.1` in your versions instead of just `.1`? Are you aware of how `/` works in versions?

`/.1` for specific_versions is not supported. Just have the version as `1.2.1.1` in specific_versions or update the docset in the `1.2.1` version and don't add a new version.

nextrevision:
The docset version, I assume, should match the upstream version of the code. If I introduce a fix to a docset, but the upstream version remains the same as before, will Dash know that an upgrade of that docset is available, or does it require a version bump in order to upgrade?

Kapeli:
Great. In that case it's fine to keep the main version as `1.2.1/.1`, but the specific_versions needs to be changed to remove the `1.2.1/.1` version completely. Just add a new docset archive for the `1.2.1` version.

There's no point in doing the `/.1` trick for specific_versions because there's no updating mechanism for specific_versions.

nextrevision:
Excellent, thanks for the explanation. I updated the PR based on your recommendations.

Kapeli:
Great. Thank you! :+1:

PR #567, opened 2015-08-01
Commit PyMel Docset

alexwidener:
PyMel 1.0.7 Initial commit


alexwidener:
I have no idea what that means. I looked at the details. Can I get an explanation?

Kapeli:
The error is: `icon.png` must be `16x16` in size. `icon@2x.png` must be `32x32` in size.

alexwidener:
Thanks, it wasn't showing that anywhere.
Should be fixed now!

Kapeli:
Looks good. One more issue: why did you include the docsets inside `versions` folders too? You don't need to do that unless you're using the `specific_versions` in docset.json (which you aren't).

alexwidener:
Because I planned on updating them. Blender updates moderately frequently and PyMel updates yearly. It's pretty common for people to be on older versions in studios, so if someone grabs it next year, they'd be caught up. Want me to remove them until then?

Kapeli:
Either remove them until then, or add the appropriate entries under `specific_versions`. It's ok to have just 1 entry inside `specific_versions`, so there's no need to hide them until then.

alexwidener:
Hi Bogdan, is there anything else holding this up or can it be merged and closed? 
Thanks!

Kapeli:
Looks good. Sorry for the wait. GitHub doesn't send notifications when you add a commit, so I didn't notice you fixed things until you commented just now.

PR #570, opened 2015-08-02
Info.plist fixed

phi1ipp:

PR #572, opened 2015-08-03
Fixed #570

phi1ipp:


Kapeli:
Few more issues:

1. Don't have the version number of the docset anywhere except for the `version` key in `docset.json`. So, remove the version number from your project folder, from the docset inside Info.plist and the docset name itself and anywhere else you might have it.
2. There's no need to have "API" in the docset name. Most docsets are about APIs, so it's ok to just use "OIM" for the docset name.
3. You don't have a `dashIndexFilePath` set. See https://kapeli.com/docsets#settingindexpage.
4. Did you generate this docset using [javadocset](https://github.com/Kapeli/javadocset#readme)? As far as I can tell, you didn't. Please consider using javadocset because you'd get a few more improvements for free (like table of contents support).

PR #574, opened 2015-08-05
Updated re #572

phi1ipp:
Can't actually use Xcode as I'm on linux. Other than that updated the docset according to your recommendations.


Kapeli:
Merged. Thank you!

PR #582, opened 2015-08-06
Better Kotlin index

sirlantis:
The current Kotlin index only has `Guides` with weird names like `stdlib / kotlin.Boolean`. This PR fixes the names and sets the correct types.

Sorry for bugging you again today. :no_mouth:


Kapeli:
Great. Thanks!

PR #584, opened 2015-08-06
Obspy docsets

megies:


megies:
Don't understand that error. The mentioned file is in the tgz..

Kapeli:
This file exists:

```
packages/autogen/obspy.core.util.geodetics.flinnengdahl.html
```

This file is mentioned in docset.dsidx:

```
packages/autogen/obspy.core.util.geodetics.FlinnEngdahl.html
```

Paths are case-sensitive, so that docsets work on case sensitive file systems too.

PR #588, opened 2015-08-09
Add GeoTools 13.2 docset

StephenRudolph:


Kapeli:
Looks good. Only one issue: you also need to add the archive for the main version (i.e. outside the versions folder). In your case, you should have the archive at this path: `docsets/GeoTools/GeoTools.tgz`.

StephenRudolph:
Thanks @Kapeli! I've now copied the main version to the docsets/GeoTools folder.

Kapeli:
Thank you! :+1:

PR #593, opened 2015-08-12
Update Dash Docset for DokuWiki Detritus release

gamma:


gamma:
Just found problem with the docset.

PR #597, opened 2015-08-18
Add GeoTrellis 0.10.0-M1 docset

zifeo:


Kapeli:
Thank you! :+1: 

Small issue: the docset doesn't have an index page. See https://kapeli.com/docsets#settingindexpage if you'd like to add one.

zifeo:
There is something weird going on with the index on Dash, that is why I did not add it. My guess is that something is not up to date when generating scaladoc with the Intellij. Basically, the side panel keeps being displayed on Dash (when showing the index, half screen) whereas other docsets do not. Any idea ? :/

Kapeli:
The index uses frames as far as I remember. The trick is to point Dash to the page that's loaded in the right-side frame, and not to the whole index page.

If it still gets redirected, look at the source for some JavaScript that does the redirecting.

zifeo:
Yep, I see, thanks! For the moment the index does not contain any essential informations so it is not mandatory. I will look for the next update :smile:.

PR #598, opened 2015-08-18
Update LLVM docset to 3.6.2

Chilledheart:


Kapeli:
Failed validation:
```
File not found at "./BitSets.html" for entry with row id "375" of "LLVM.docset"
File not found at "./CommandGuide/llvm-lib.html" for entry with row id "729" of "LLVM.docset"
File not found at "./BuildingLLVMWithAutotools.html" for entry with row id "780" of "LLVM.docset"
```

This means that you have entries in your docSet.dsidx which are not valid or point to files that don't exist. I think this happens because of the `./` at the beginning of the path.

PR #610, opened 2015-08-26
fis3 Docsets

ah-kevin:
Initial Commit


Kapeli:
You didn't add the `fis3.tgz` at the right spot. You're not supposed to put it in `/versions/` if you don't use `specific_versions`. Just add it at the root of your folder.

Kapeli:
Also, only English docs are allowed.

PR #620, opened 2015-09-03
Netty5 docsets

zhouxiaoxiaoxujian:
The docsets for Netty5


Kapeli:
Issues:
1. Just call it "Netty", not "Netty5" or "Netty5 Docset"
2. You need to add the archive in the main folder, not just in the `versions/` folder
3. The icons you added are not `16x16` and `32x32`.

PR #639, opened 2015-09-13
added electron 0.32.0 docset

pirafrank:
I've deleted my fork and reforked your repo.
What was wrong?


Kapeli:
Thank you! :+1: 

> What was wrong?

Whenever I grab docsets, I rewrite the git history to remove the docsets from the git repo (to keep its size small). You should never pull changes from this repo into your fork.

pirafrank:
Oh ok, sorry about that. Thank you for your reply!

PR #670, opened 2015-10-08
Adding note on how to use user contributed docset

tianhuil:


Kapeli:
I don't think this is needed. Sorry.

PR #672, opened 2015-10-10
updated to latest otrs 3.2, 3.3., 4.0 and added OTRS beta 5

hanneshal:


Kapeli:
What is `docsets/OTRS/versions/otrs32.tgz` for? It's not mentioned in the `docset.json`.

hanneshal:
It should be in there. In my latest commit, which is the same as my PR it is: 
https://github.com/hanneshal/Dash-User-Contributions/blob/master/docsets/OTRS/docset.json#L28

hanneshal:
Ah, now I get it. The folder is missing. I recreate the PR

PR #694, opened 2015-11-04
Update Parse iOS SDK docset to version 1.9.1

AJMiller:


Kapeli:
`Parse-iOS-SDK.tgz` needs to be in the main folder as well (next to `docset.json`), not just in the `versions` folder.

AJMiller:
Ok added to root Parse_iOS_SDK directory. Let me know if there are any other changes needed!

Kapeli:
Thank you! :+1:

PR #701, opened 2015-11-13
updated Kotlin docset to 1.0.0-beta-1103

sirlantis:
Updated very outdated Kotlin docset (they changed a lot for 1.0-beta). Does the following cause an issue?

    "version": "1.0.0-beta-1103"

Or should I set it to `1.0.0/beta-1103`, which would be a little bit confusing I guess.


Kapeli:
I think `1.0.0-beta-1103` is fine. Thank you! :+1:

PR #702, opened 2015-11-14
a new docset, SPL, Splunk Search Processing Language

jovezhong:
Splunk is a popular big data tool. It also provides free edition for log analysis. Search Processing Language, or SPL, is its language to effectively transform and analysis data.


Kapeli:
The entries you have for Functions and Methods link to the top of the pages instead of anchors within the pages where they are documented.

PR #708, opened 2015-11-22
Added Ecto docset

hideshi:
I've just added a docset of Ecto, which is a major ORM library of Elixir language that is integrated into Phoenix framework.


Kapeli:
This docset is already available in Dash in Preferences > Downloads > Hex Docsets.

PR #716, opened 2015-11-28
Added async.js docset

mgill25:


Kapeli:
This is a cheat sheet. You should contribute it here: https://github.com/Kapeli/cheatsheets.

PR #718, opened 2015-11-30
Updating Jenkins Job Builder to 1.3.0

nextrevision:


Kapeli:
Latest versions need to be at the top, not at the bottom (i.e. you should have `1.3.0` before `1.2.1`).

nextrevision:
I see, thanks for pointing that out. Should be fixed now!

Kapeli:
Thank you! :+1:

PR #725, opened 2015-12-04
Add WPILib Java Docset

michaelsilver:
[WPILib](http://wp.wpi.edu/wpilib/) is the library used wordwide by thousands of people to program robots for the [FIRST Robotics Competition](http://www.firstinspires.org/robotics/frc). I generated this docset using [javadocset](https://github.com/Kapeli/javadocset), and then added logos.

This docset contains the current version of the documentation for the 2015 release (note how I labeled the version). I was not sure whether I should write something in the `specific_versions` part of the `docset.json` since this is the first version, so please let me know if I need to add anything there. 

Thank you!


Kapeli:
The `specific_versions` should be this:

```json
"specific_versions": [
    { 
        "version": "1.0.0",
        "archive": "versions/1.0.0/WPILib.tgz"
    }
 ]
```

michaelsilver:
Ok, I added version 1.0.0 to the `specific_versions`. Can the version number in `specific_versions` be `1.0.0/2015` instead of just `1.0.0` (to be consistent with `"version": "1.0.0/2015"`)?

I also added a few appropriate aliases.

Kapeli:
Thank you! :+1: 

> Can the version number in specific_versions be 1.0.0/2015 instead of just 1.0.0 (to be consistent with "version": "1.0.0/2015")?

Nope. The `/` trick is so that you can trigger a docset update without changing the version the users see (i.e. `1.0.0/blabla` and `1.0.0/2015` will both show up as only `1.0.0` to users).

Specific versions never update, so if you add the `/2015`, it will be shown to the users.

PR #726, opened 2015-12-04
Processed docsets/Alembic/Alembic.tgz

asheraScout:


Kapeli:
Please change the icon or remove them completely.

asheraScout:
OK, I'll create another commit

PR #734, opened 2015-12-11
Adding Editorial & Pythonista docsets for the iOS apps.

leesavide:
As these contain changes that are specific to the iOS architecture, the Pythonista docset differs from the standard Python documentation, NumPy documentation, Matplotlib documentation, and SymPy documentation. Likewise, Editorial also differs from the standard Python documentation, and also includes modules for custom workflows and logic specific to it's context as an iOS app.


Kapeli:
Please redo your fork and resubmit.

Kapeli:
Also an issue: the author inside `docset.json` needs to be the author of the docset, not of the docs themselves. So that should be you, not Ole.

leesavide:
You got it.

PR #737, opened 2015-12-11
[Reissue] Adding Editorial & Pythonista docsets for the iOS apps

leesavide:
Resubmitting pull request.

As these contain changes that are specific to the iOS architecture, the Pythonista docset differs from the standard Python documentation, NumPy documentation, Matplotlib documentation, and SymPy documentation. Likewise, Editorial also differs from the standard Python documentation, and also includes modules for custom workflows and logic specific to it's context as an iOS app.

Signed-off-by: Lee Savide <laughingman182@gmail.com>


Kapeli:
Thank you! :+1:

leesavide:
☜(●▽●☜) Anytime!

Also: I did have to add a .gitattributes file to allow for git-lfs to push the Pythonista docsets. It might be a good idea to have anyone pushing a docset over 50 MB to use git-lfs as well.

Kapeli:
Ah. I thought the .gitattributes are not needed. For some reason it causes an error when pulling on OS X (`git: 'lfs' is not a git command`).

leesavide:
On Mac OS X the command gets named as git-lfs, but then needs to have aliases set to allow for `git lfs <options>` to work. I had the same problem yesterday when working with other repos.

The gitattributes I added here are only needed for specifying which files use LFS. Since Pythonista 1.6 is in beta, I figured that it'd avoid complications early and ensure that it's documentation could be updated quickly. It only affects the Pythonista & Editorial archives, but for future use case, it might be easier to add `*.tgz filter=lfs diff=lfs merge=lfs -text` to the gitattributes, so anyone else contributing can upload a docset of any size, or to tell users should set this in their global gitattributes file.

Kapeli:
I don't think making an alias and so on should be a requirement for this repo, so I'm against the `*.tgz` bit. I'll leave things as-is for now. Feel free to re-add the .gitattributes when you update the docset next time.

leesavide:
Ah ok. :) Thanks a bunch!

PR #738, opened 2015-12-11
Added electron docset v0.35.0

wearhere:
My impression is that I need to add the latest version both at the root of the docset folder and in the specific version folder, hope that was right! cc @pirafrank


Kapeli:
Yep, that's right.

@pirafrank I've verified this and I'm 99.99% it's ok to merge so I'll go ahead and do that. If anything is wrong let me know.

pirafrank:
Hi Bogdan,

thanks for pinging me and letting me know.
Nothing wrong :)


Francesco Pira
fpira.com

PR #740, opened 2015-12-12
added electron docset v0.34.0

pirafrank:
just notices docset for version 0.34.0 was missing. added for completeness.


Kapeli:
Thank you! :+1:

PR #741, opened 2015-12-12
Update D to 2.069.2

trishume:
Dlang has been changing its standard library a bit in the last couple versions. I just ran my script again and scraped the new site.

@Kapeli


Kapeli:
Thank you! :+1:

PR #755, opened 2016-01-07
Upgrade to Bokeh 0.11.0

jbcrail:
I modified the stock release so that the external stylesheet was embedded and the version number was hardcoded (instead of using the internal library version).


Kapeli:
Thank you! :+1:

PR #758, opened 2016-01-07
Async.js docset

silentroach:
Popular Node.js package [Async.js](https://github.com/caolan/async) docset.


Kapeli:
Thank you! :+1:

Issue #754, opened 2016-01-07
Keyword for Eigen


ppwwyyxx:
According to https://github.com/zealdocs/zeal/issues/383, the default search keyword for docset is ``DocSetPlatformFamily``.
With the eigen docset, this keyword is set to ``doxygen``, which doesn't make a lot of sense.

Kapeli:
@elrnv Can you set the keyword to something like `eigen`?

elrnv:
Can do. It must have been changed during the latex pictures were updated, so I'm gonna need some time to figure out how to get a build with latex pictures to show properly ( would appreciate a hint from @oracleyue :) ). Also, I might as well update the docs to the latest version.

PR #759, opened 2016-01-08
Added docs for Eigen version 3.2.7

elrnv:
Updated the Info.plist file in both versions to have "eigen" as the dash keyword.
Updated the README.md and docset.json to reflect the version update.
NOTE: The binary in the Eigen directory is the same as what's in versions/3.2.7/

### DIFF
--- a/docsets/Eigen/Eigen.tgz.txt
+++ b/docsets/Eigen/Eigen.tgz.txt
@@ -1,4 +1,0 @@
-Archive "Eigen.tgz" was processed at this location, pushed to the CDN and completely removed from git.
-
-Date: 2015-06-09 20:15:02 +0000
-SHA1: 8678f87b28690c1bcbbb663c2a1845e34a51c448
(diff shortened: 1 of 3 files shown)


Kapeli:
There's no point in having docsets in the `versions` if you don't have a `specific_versions` in your `docset.json` (https://github.com/Kapeli/Dash-User-Contributions/blob/master/Sample_Docset/docset.json#L14-L26). Either add the `specific_versions` or remove the docsets from `versions`.

elrnv:
I'm not sure why the build failed here. It seems like it failed at the 3.2.4 version. I copied that docset directly from the current Eigen docset (what's currently on Dash), and only changed a few things in the Info.plist file. It works fine when I load it onto Dash from my machine.
Any hints on what I can do to make it work?

Kapeli:
The 3.2.4 docset is an archived docset that was re-archived by you. Look in the Resources folder and you'll see there are no HTML files there. This means that this docset was installed by Dash at some point and you copied it over. You shouldn't do that.

Kapeli:
Thank you! :+1:

elrnv:
My pleasure! BTW this resolves #754 :)

PR #762, opened 2016-01-09
Initialize an Apache Hive docset with validation issues fixed.

wezhang:


wezhang:
Ok...the file name 'info.plist' doesn't meet the requirement.

PR #763, opened 2016-01-09
Initialize an Apache Hive v1.2.0 docset with info.plist filename issues fixed.

wezhang:


wezhang:
```
Empty name found for row id "1011" at "Apache_Hive.docset"
```

PR #770, opened 2016-01-12
Update WPILib to 2016 version

michaelsilver:
Additionally, I have added comments in `specific_versions` to clarify what each version means, because the maintainers of WPILib do not have official versions, only yearly releases corresponding to the year. 

It does occur to me that a potentially better way to create "fake" version numbers is to use `2.0.15` and `2.0.16`, etc. (for the years). @Kapeli would you prefer this, and if so, could I just change `1.0.0` to `2.0.15` in `docset.json` and `versions/` or would that mess it up for current users of the docset?


Kapeli:
Why not set the versions to 2015 and 2016? The version can be anything. It doesn't have to be a real version.

michaelsilver:
Great, that's even better! Didn't know that any number is allowed, not just semantic versioning. I've updated the version numbers to the years as you suggested.

Kapeli:
Thank you! :+1:

PR #775, opened 2016-01-13
Update to Spark 1.6

ronert:


Kapeli:
There's no point in adding `versions/1.5.1/pyspark.tgz` without also adding the required entries in `docset.json` ([like this](https://github.com/Kapeli/Dash-User-Contributions/blob/master/Sample_Docset/docset.json#L14-L25)). Either remove the archive or fix the `docset.json`.

ronert:
Thanks, done.

Kapeli:
Thank you! :+1:

PR #779, opened 2016-01-16
Add Re:VIEW docset

orangain:
[Re:VIEW](https://github.com/kmuto/review) is an easy-to-use digital publishing system for books and ebooks.


Kapeli:
Thank you! :+1:

PR #780, opened 2016-01-16
Add first update to 2016 WPILib javadocs

michaelsilver:
I want to make users recieve this update without changing the version number. I followed the [Version Guidelines](https://github.com/Kapeli/Dash-User-Contributions/wiki/Docset-Versioning-Guidelines) as best as I could understand, but please let me know if need to change anything.


Kapeli:
The `/update1` is enough to trigger an update and the version that users will see will be just `2016`..

PR #792, opened 2016-01-26
Updated docset to 0.5 prior to hrev50000

return:
Haiku Book docset update to 0.5 due to hrev50000.


Kapeli:
You need to add `Haiku_Book/Haiku_Book.tgz` too (use the same file as `Haiku_Book/versions/0.5/Haiku_Book.tgz`), otherwise you're not updating the main docset you're only updating the versioned one.

return:
The main docset has now been added to the repository.

Kapeli:
Thank you! :+1:

Issue #791, opened 2016-01-25
Request: IntelliJ Platform SDK


dubuqingfeng:
The documentation is here: http://www.jetbrains.org/intellij/sdk/docs/index.html
The repository is here: https://github.com/JetBrains/intellij-sdk-docs
ths

PR #799, opened 2016-01-28
add Intellij Platform SDK Docset

dubuqingfeng:
About #791 
Thanks.


Kapeli:
Please rename the docset (in `Info.plist` and `docset.json`) to "IntelliJ Platform SDK" instead of "Intellij_Platform_SDK".

Also please index the docset properly. The docset index has only 5 entries.

PR #800, opened 2016-01-28
add Intellij Platform SDK Docset

dubuqingfeng:
about #791 
Thanks for your help.


Kapeli:
The name should be "IntelliJ" (uppercase J), not "Intellij". Also please redo your fork and squash all commits into one.

PR #808, opened 2016-01-31
added updated electron docsets

pirafrank:
added electron docset versions 0.36.3, 0.36.4 and 0.36.5 (latest)


pirafrank:
I think it should be ok now :)

Kapeli:
Thank you! :+1:

PR #821, opened 2016-02-07
Update Vaadin to version 7.6.2

ewall:
Updated as requested, looks like bad timing on the previous pull request.


Kapeli:
Thank you! :+1:

PR #827, opened 2016-02-13
Add second released update to 2016 WPILib javadocs

michaelsilver:
This should trigger an update for users without changing the version number.


Kapeli:
Thank you! :+1:

PR #836, opened 2016-02-23
ADD new version to npm (npm 3.7.3)

francoismassart:
Not certain about what is the `npm.docset` file at the root of the folder (same level as the `versions` folder)... It is a copy of the latest docset (same file as `versions/3.7.3/npm.docset`)... Archived too!
Is this correct?


Kapeli:
Failed validation:
```
Could not find docSet.dsidx at "docsets/npm/versions/3.7.3/unarchive_temp/npm.docset/Contents/Resources/docSet.dsidx"
```

Also please redo your fork and squash all commits into one.

PR #850, opened 2016-03-03
wxWidgets 3.1.0 docset

davidwed:


Kapeli:
You shouldn't place the docset archive within the `/versions/` folder if you don't have `specific_versions` within your docset.json. In your case, you should just put `wxWidgets.tgz` at this path: `docsets/WxWidgets/wxWidgets.tgz`.

Also the `docSet.dsidx` seems to be corrupt. I can't load it in Dash.

PR #857, opened 2016-03-15
Add third released update to 2016 WPILib javadocs

michaelsilver:
This new version of the docs was released on March 1st. (Sorry for not catching it sooner!)

These changes are intended to trigger an update for users without changing the version number, as it is still the 2016 release.


Kapeli:
Thank you! :+1:

PR #868, opened 2016-03-21
update phaser to 2.4.6

nisimjoseph:


Kapeli:
You need to rename the archive to `.tgz`. Don't use `.tgz.txt`. You also need  to update the `docset.json`.

@rcolinray Any chance you could update the Phaser docset yourself?

nisimjoseph:
fixed

PR #878, opened 2016-03-27
update tldr pages to version 0.0.2.15.

Moddus:
update tldr pages to version 0.0.2.15.


Kapeli:
The `Info.plist` inside the docset is not valid. It's missing a `<` at the beginning.

PR #881, opened 2016-03-28
Phaser 2.4.6 docset update

nisimjoseph:


docsets/Phaser/docset.json
```diff
@@ -1,10 +1,10 @@
 {
     "name": "Phaser",
-    "version": "2.4.3",
+    "version": "2.4.6",
     "archive": "Phaser.tgz",
```
Kapeli: This points to `Phaser.tgz`. The file you added is called `phaser.tgz`.

PR #885, opened 2016-04-03
Basscss initial

VictorHom:
I would like upload bass css docs. Also thanks for the help thus far.


VictorHom:
looks like I failed the travis tests. I'm gonna look at it later, any suggestions on what it is looking for? I see that it says missing files.

Kapeli:
`https://` entries are not allowed in user contributed docsets. All entries should be local.

PR #910, opened 2016-04-21
Updates diesel docs to version 0.6.1

zamith:


Kapeli:
The docset inside `diesel.tgz` is empty.

PR #915, opened 2016-04-23
Adds docset for Robot Framework 3.0

deiga:


Kapeli:
You need to add the docset inside the `docsets` folder, not inside the root folder of the repo. Also, you need to have an `icon@2x.png` (32x32 in size), or not have an icon at all.

PR #937, opened 2016-05-22
Add FileMaker 15 Docset

jpscharf:


docsets/FileMaker/docset.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "FileMaker",
-    "version": "14.0.1",
+    "version": "15.0.0",
     "archive": "FileMaker.tgz",
```
Kapeli: You need to add the docset archive at this path too, not just in the `versions` folder.

jpscharf: Fixed, sorry about that.

Kapeli:
Thank you! 👍

PR #943, opened 2016-05-28
update seaborn to v0.7.0

gouthamvel:


Kapeli:
Please change the docset name to "Seaborn" and the docset archive to "Seaborn.tgz".

Kapeli:
You need to change the name of the docset too (inside Info.plist).

PR #949, opened 2016-06-03
Update Three.js to r77

drewlustro:
Now that documentation has been released for r77, I've regenerated the docs.


Kapeli:
Thank you! 👍

PR #954, opened 2016-06-07
Add Brunch 2.8.0 docs

benrki:


Kapeli:
`icon@2x.png` is not 32x32.

PR #957, opened 2016-06-09
Cinder docset update

naychrist:
Hi there, finally got around to updating the Cinder docset for the latest Cinder. Thanks again for the great app!


Kapeli:
Thank you! 👍

PR #958, opened 2016-06-09
added all missing docsets from 0.36.6 onward (till current 1.2.1))

pirafrank:
Please note that the following patch versions are missing because they never had the corresponding documentation on the website.

- 0.36.12
- 0.37.8
- 1.0.0
- 1.0.2


Kapeli:
Thank you! 👍

PR #959, opened 2016-06-09
TensorFlow Docset

ppwwyyxx:
I'm transferring my docset at https://github.com/ppwwyyxx/dash-docset-tensorflow to here, so more people can use it.
My first time to contribute a docset. Hope the format is right.


Kapeli:
Thank you! 👍

PR #960, opened 2016-06-10
update boto3 docset to version 1.3.1

mtrbean:


Kapeli:
Thank you! @shanedroid if anything is wrong, let me know and I'll revert the update.

PR #961, opened 2016-06-10
Added CAPI docset

fourier:
Added Lispworks CAPI docset: 
http://www.lispworks.com/documentation/lw70/CAPI-M/html/capi-m.htm


Kapeli:
Thank you! 👍

PR #965, opened 2016-06-14
Update Pythonista 3 docset & add legacy Pythonista 2.1 docset

leesavide:
Updated docsets to Pythonista 2.1 for legacy version, & Pythonista 3.0 for
current version. Also reduced the docset size, because Pythonista
3.0's documentation includes a large number of 3rd party modules 
that Dash already has specific docsets for, and it would be very time
consuming to attempt to include all of them into one docset package.


Kapeli:
Thank you! 👍 

I've noticed one issue: versions under `specific_versions` need to be sorted by latest/newest first. I've fixed this in https://github.com/Kapeli/Dash-User-Contributions/commit/04925958c5609078d7e9f222f2290f293c77f3b1.

PR #975, opened 2016-06-23
change folder name to loopback which is same with the docset name

yanqiw:


yanqiw:
ops... still not pass the validation, could you let me know what's going on, and reject this pull request? Thx

Kapeli:
The issue is the same as before:

```
Found entry with new line in its name at row id "3" at "loopback.docset". Complete entry name is: "Version 2.29.0
        
          (latest)"
```

This means that one of the entries in your docSet.dsidx contains a new line.

PR #976, opened 2016-06-23
remove sample docs for validation again.

yanqiw:


yanqiw:
well, I don't really know where is wrong, could you let me know how to change the doceset?

PR #984, opened 2016-06-27
rename scikit to scikit-learn

sciunto:
I fixed the path for scikit-learn. Thanks.


Kapeli:
This would cause Dash to treat this docset as a new one and all users that have the docset installed won't get updates anymore. Sorry.

sciunto:
That's sad, the naming is just wrong. :(

Kapeli:
The naming only affects this repo and not end-users, so it's fine.

sciunto:
Not really. I'm discovering all the work around dask, but for instance, here: https://github.com/zsoltika/feeds the xml file has a wrong name.

PR #985, opened 2016-06-28
Update astropy docset to v1.2

deprecated:
These haven't been updated in a while, so I took the liberty of trying it myself.  The original author @iamaziz pulled the html docs from readthedocs and then ran doc2dash on them.  But I couldn't get that to work, so I built the docs from the source repo instead.


Kapeli:
Thank you! 👍

PR #986, opened 2016-06-29
Fixed Pythonista 1.5 docset to match indexing from current versions.

leesavide:
If needed, the tarball is also available from [here](https://drive.google.com/open?id=0B0dGDuzxiYinQm1lcG1qR2JOaHc).


Kapeli:
Thank you! 👍

PR #988, opened 2016-07-01
Added updated docsets for Blender API 2.77.1 

alexwidener:
There's an extra commit from an old fork and clone, should be good now as far as I can tell. Sorry about that. 

Thanks!


Kapeli:
The `specific_docsets` need to be sorted by latest-first.

You also need to add the archive to the main folder and update the main version inside the docset.json.

Kapeli:
`docset.json` points to `Blender.tgz`. You added `blender.tgz` (case sensitive).

Kapeli:
Please also delete the archives with the lowercase name.

Kapeli:
Thank you! 👍

PR #997, opened 2016-07-12
Update Pure CSS docs

jkrehm:
Updates from 0.5.0 to 0.6.0


Kapeli:
If you want to add an icon, it has to be `16x16` and also you need to have an `icon@2x.png` which would have to be `32x32`.

Also please fix all the CSS issues inside the docset.

@alexlafroscia Are you ok with @jkrehm taking over the maintenance for this docset?

jkrehm:
Sorry, I didn't know how to test the docset before uploading, but I assumed at least the HTML was fine. I was wrong on that account. Pure uses some server-side requests to get specific CSS files, and SiteSucker doesn't know what to do with them.

I thought this would be a fairly quick update, but it's turning out to be a pain. No wonder Alex hasn't updated it in 2 years.

I'm going to close this for now. Perhaps if I have a few hours to muck about fixing up stuff manually I'll create a new PR.

alexlafroscia:
Yes please 😁

PR #1009, opened 2016-07-21
Update Scikit-learn 0.16.1 to 0.17.1

iamaziz:
@Kapeli I updated the folder name too, to match the docset.


Kapeli:
The folder name is used as the unique identifier within Dash. If I accept this pull request then everyone that has this docset installed would be stuck with a version that doesn't update because it would use (and look for updates) using the old identifier/folder name.

In my opinion, this isn't worth it and we can just keep the folder name as it is.

iamaziz:
Fair enough, what about the update though ? accept it then revert the folder name to the old one ?

Kapeli:
Sorry, didn't notice there's an update too. Can you submit it in a separate pull request please? I merge this one, but it's kind of hard to remove `tgz` files from git history after a rename.

iamaziz:
Sure thing, already submitted a brand new pull #1010 with the legacy folder name :-)

PR #1018, opened 2016-07-29
Add new docsets for Salesforce Apex and Visualforce

ViViDboarder:
Two new languages used in Salesforce. One for controllers and the other for markup.

One thing I'm not sure I did correctly is the aliases.


Kapeli:
The aliases are ok. Thank you! 👍

PR #1023, opened 2016-08-03
Update Terraform docset to 0.7.0

f440:


Kapeli:
Failed validation:
```
Found directory in main index: Entry with row id "17" and path "docs/commands/state" at "Terraform.docset"
Checking Terraform/versions/0.7.0/Terraform.tgz...
Found directory in main index: Entry with row id "17" and path "docs/commands/state" at "Terraform.docset"
```

This means you have an entry in your `docSet.dsidx` that points to a folder.

PR #1027, opened 2016-08-11
Update FFmpeg docset to 3.1.2

klausbadelt:


Kapeli:
You should not include the docset in the pull request, just an archive of it. Please redo it.

PR #1037, opened 2016-08-17
Add Standard ML Docset

Luminoid:


Kapeli:
You need to set the archive to "Standard ML.tgz" in the docset.json as well.

Kapeli:
Thank you! 👍

PR #1038, opened 2016-08-17
Update Pillow docset to version 3.3.0

153957:


Kapeli:
Moving the text file to `docsets/Pillow/versions/2.8.2/Pillow.tgz.txt` doesn't work. You need to add the actual `tgz` at that location.

153957:
Ok, in that case I will simply remove it, or is it important to keep old versions around?

Kapeli:
I have no idea really. Whether or not old versions are needed is up to you. Are the docs significantly different and/or it's common for Pillow developers to stay on the old version?

153957:
So when I would for instance release an update for version 4 and I would like to keep the v3.3 docs I should each keep the tgz around or regenerate it in order to put it in the specific_versions/archive? So the SHA1 is not used as a hash to find the file on the server to download? Only to check if it is the correct one?

Kapeli:
The SHA1 is not used for anything, it's just something my script leaves behind to let you know I processed the tgz.

You can avoid keeping a copy of the tgz by downloading it from my server when you need to add it to specific_versions: https://kapeli.com/feeds/zzz/user_contributed/build/Pillow/Pillow.tgz.

153957:
Ok thanks, I added the tgz of the previous version.

Kapeli:
Thank you! 👍

PR #1044, opened 2016-08-21
Updated Chrome API docset to version 46

gebrkn:


Kapeli:
The docset's docSet.dsidx seems to be incomplete/malformed. There's a SQLite journal left behind so maybe you forgot to commit the SQL transaction? Please redo.

PR #1050, opened 2016-08-25
Add Kirby 2.3.2

samnabi:


docsets/Kirby/docset.json
```diff
@@ -12,6 +12,10 @@
         { 
             "version": "2.1.0",
             "archive": "versions/2.1.0/Kirby.tgz",
+        },
+        { 
+            "version": "2.3.2",
```
Kapeli: Specific versions need to be ordered by latest-first, so `2.3.2` should come before `2.1.0`.

PR #1051, opened 2016-08-25
update RFCs docset to v1.3

willnorris:
This docset is over the 100 mb limit, so I've not included the tarball in this commit. The current 1.3 version can be obtained from https://github.com/willnorris/rfcdash/releases/download/v1.3/RFCs.tgz

### DIFF
--- a/docsets/RFCs/docset.json
+++ b/docsets/RFCs/docset.json
@@ -1,6 +1,6 @@
 {
     "name": "RFCs",
-    "version": "1.2",
+    "version": "1.3",
     "archive": "RFCs.tgz",
     "author": {
         "name": "Will Norris",


Kapeli:
The docset fails validation:

```
File not found at "tools.ietf.org/html/index.html" for entry with row id "3431" of "RFCs.docset"
```

This means that the entry with row id `3431` points to a file that does not exist.

willnorris:
ah interesting.  What are you doing to validate the docset?  Is there a tool I overlooked somewhere? I just tried manually loading it and it seemed fine, but didn't notice this.

Kapeli:
The check is usually made by Travis when you open the pull request, but as the docset is over 100 MB this doesn't happen.

You can make the check yourself by running this command:

```
wget https://kapeli.com/feeds/zzz/docsetcontrib.tgz && tar -xzf docsetcontrib.tgz && ./docsetcontrib --verify
```

You need to be inside the "Dash-User-Contributions" folder (the root of the git repo) when you run the command. Also the docset archive ("RFCs.tgz") has to be at "Dash-User-Contributions/docsets/RFCs/RFCs.tgz".

Kapeli:
Also you can check the error within Dash by searching for "IETF Documents", as this is the entry with the wrong path.

screen shot 2016-08-26 at 10 26 02

PR #1053, opened 2016-08-26
Add Odoo 8 docset

naglis:
Hello,

this adds [Odoo v8.0](https://www.odoo.com/documentation/8.0/) docset. This is my first generated docset, so please let me know if something is not done correctly.

Thanks!


Kapeli:
Thank you! 👍 

One thing I've noticed: when you select an entry in Dash the docs do a weird scroll animation. The javascript that causes this should be removed from the docset.

PR #1055, opened 2016-08-28
update OpenResty to 1.11.2

spacewander:


Kapeli:
Thank you! 👍 

One note: in `specific_versions` the latest/newest versions should be at the top. I'll fix it.

PR #1081, opened 2016-09-17
Add docset for toyplot

ProgramFan:
[Toyplot](http://toyplot.readthedocs.io) is an interactive plotting package for Python. This PR adds toyplot docset for Dash.


Kapeli:
The name and keyword of the docset (in `Info.plist`), should be "toyplot" and not "html".

PR #1082, opened 2016-09-17
Add toyplot docset (name fixed)

ProgramFan:
Toyplot is an interactive plotting package for Python. This PR adds toyplot docset for Dash.


Kapeli:
Thank you! 👍

PR #1083, opened 2016-09-17
Fix index page in toyplot docset

ProgramFan:
Fix index page in toyplot docset. Wonder why 'index.html' is not the default.


Kapeli:
Please redo your fork and resubmit.

PR #1096, opened 2016-09-25
Added ns-3 docset.

tnull:
I added a docset for the [ns-3](https://www.nsnam.org/) network simulator. 
However, the `ns-3.tgz` is around 2.5 GB, so I'd need somewhere to upload it before the PR can be accepted.

Regards!


tnull:
I uploaded the `ns-3.tgz`, you can pull it from: https://tnull.de/static/ns-3.tgz

Kapeli:
2.5 GB is a bit too large unfortunately. Any idea what's causing the size? The docset has around 2400 classes, so it shouldn't be quite that large.

tnull:
I think the size comes mostly from huge number of `png` files. These should be inheritance graphs and collaboration diagrams generated by `doxygen`/`dot`. Should I exclude them? Then, however, the documentary might be broken at some points (as in you would see that there is something missing).

Kapeli:
Maybe configure Doxygen so it doesn't generate graphs at all?

tnull:
Yep, I disabled what I could find and was able to push the file size of `ns-3.tgz` to 72 megabytes. See the  updated commits! :)

Kapeli:
Great. Please fix these issues though:
```
Not 16x16: "icon.png" in "ns-3"
Not 32x32: "icon@2x.png" in "ns-3"
```

tnull:
Done. Strange, each of the icons were exactly twice as large. Must have done sth. wrong while saving them.

Kapeli:
Thank you! 👍

PR #1115, opened 2016-10-08
Update from version 0.17.1 to 0.18.0

serhanaya:
[Fixed] Added Scikit-learn.tgz

Update to the new version of scikit-learn (v.0.18).

P.S Line 25 of the file scikit-to-dash.py needs to be changed from
urllib.urlretrieve to urllib.request.urlretrieve (I am waiting for my
pull request to be accepted by Aziz Alto).

PR #1116, opened 2016-10-09
Documentation for scikit-learn v0.18.0

jaypeedevlin:
This replaces the outdated scikit-learn docs


jaypeedevlin:
I see that there is an existing PR for these docs [here](https://github.com/Kapeli/Dash-User-Contributions/pull/1115)  I'm happy for mine to be closed, or if mine passes CI (I see it currently doesn't) mine to be used.

serhanaya:
Hi Josh! You absolutely made a better version than mine, also improved the original repo. I’m a bit ashamed about the fails because this is my first PR :) I’ll be looking the details of your PR, but I need to wake up first, it’s 5 am here.

Kapeli:
Looks good, but please move everything back into the "Scikit" folder. Dash uses the folder name as the unique identifier for a docset, so if the folder is renamed then Dash will treat the docset as a new one and existing users that have Scikit-learn installed won't get the update.

The folder name isn't displayed anywhere to users, so even though it's wrong, it's better to leave it as-is then to cut off users from updated.

jaypeedevlin:
@Kapeli fixed folder name

Kapeli:
Thank you! 👍

PR #1118, opened 2016-10-10
Full API Document for ffmpeg/libav

steve3d:
current docset for ffmpeg is almost useless without api documents.

here is the full api documents set for ffmpeg and libav, also includes the original guides.


Kapeli:
Failed validation. See https://travis-ci.org/Kapeli/Dash-User-Contributions/builds/166496713 for log.

PR #1121, opened 2016-10-11
Full API Document for FFmpeg/libav

steve3d:
here is the full api document for ffmpeg/libav.

fixed the missing pages in index.


steve3d:
well, failed again....

steve3d:
I've already fixed those errors, Should I open a new PR?

PR #1122, opened 2016-10-11
ffmpeg and Libav API documents 

steve3d:
Sorry for the tittle
The ffmpeg and libav API document


steve3d:
Ok. Failed again, But the missing files exists, I've double checked.  Don't know why checks failed.

steve3d:
All those missing files also exists in my previous PR.

Kapeli:
The paths are case sensitive.

For example in the index you have:
```
api/structNuvContext.html
```

While the actual file is at:
```
api/structNUVContext.html
```

Also, `Libav/Libav.tgz` should be an actual tgz, not a symlink to the one found in versions.

steve3d:
OK, got it, linux filesystem is case sensitive. and Libav/Libav.tgz should be the newest version of docset right?

Kapeli:
Yes.

steve3d:
OK then, seems building the docset under mac is not a good idea, I'll try to build it under linux, which has case sensitive fs.

steve3d:
@Kapeli, I finally found the problem, those missing files are caused by typo in source files.

Now I have two solution, 
- filter those typo symbols in source file to make it point to correct files. But most of these symbols are not exposed by the library, so it's only value is just a reference. but these might change over different versions.
- skip build all source files, so the docset only includes exported api from the library.

which should I do?

Kapeli:
When you're adding the index entry, can you convert the path to the case-sensitive one?

steve3d:
like these two
```api/structNuvContext.html
api/structNUVContext.html```
the real symbol is NuvContext, under mac, the first generated file will survive, but under linux, I got both file exists: they both exists.

steve3d:
@Kapeli , I've made some research, build a typo symbol list to skip is not possible, so look at the error log of my first PR, for example , the RiceContext and RICEContext, ffmpeg and libav both use them.

So the only way to do this is building these docsets on a case-sensitive filesystem. Or build these docsets without doxygen source files. for example I can create two version of docset for ffmpeg 3.1.4, one include all documents, another only includes the documents for header files.

Which one is better?

Kapeli:
I don't know, sorry.

PR #1124, opened 2016-10-11
added docs from 1.3.3 to 1.4.3

pirafrank:
ps. I've experimented the way my server script works. Docset updates should be automated/faster now.


Kapeli:
Thank you! 👍

PR #1126, opened 2016-10-12
Added SpigotMC 1.10.2 Docset

crocbuzz:
Added Sample_Docset back


Kapeli:
The archive you added is called `SpigotMC.tgz`, while `docset.json` points to `SpigotMC-1.10.2.tgz` instead.

Also, please don't include any kind of versioning in the docset name.

PR #1127, opened 2016-10-12
Adds Docset for path.py

### LINKED ISSUE #442: Add docset: path.py
I want to add a docset for the Python module `Path.py`.  

I started to follow your instructions, but your `doc2dash` tool does not work for online resources.  

Here are my requests for the new docset:
- Source: http://pythonhosted.org/path.py/api.html
- That is the only page worth adding to Dash.  (There’s a homepage that contains a link to the API page, and an auto-generated index page, but that’s it.)
- **Note:** The project name is “_Path.py_”.  However, when using it in code, the typical module usage is `from path import Path` (no _.py_ suffix).

rspeed:
Fixes #442

The path.py project doesn't seem to have a logo of any sort, so there's no icon. Also included is a build script and docfiles for the three previous major versions.


Kapeli:
Thank you! 👍

PR #1131, opened 2016-10-13
First Time commit ScalaFX docset version ScalaFX API 8.0.102-R12

haleychin:
ScalaFX is a simpler way to use JavaFX from Scala. First time upload to User Contribution.


Kapeli:
Thank you! 👍

PR #1132, opened 2016-10-13
full api docs for ffmpeg/libav

steve3d:
this time should pass the test, your test has a fatal flaw, you are expecting case-sensitive file results on a case-insensitive filesystem. That's not possible.

So I've manually excluded those symbols.


Kapeli:
> this time should pass the test, your test has a fatal flaw, you are expecting case-sensitive file results on a case-insensitive filesystem. That's not possible.

It is, if you don't have duplicate files (case-insensitive-wise).

On Mac, most file systems are case-insensitive. On iOS, the file system is case-sensitive. That's why I need to check for both cases.

Kapeli:
All good. Thank you! 👍

PR #1133, opened 2016-10-14
IonIcons Docsets

bbcdanang:
I make this docset because its not available yet in docset library. Very useful feature from android-iconify (https://github.com/JoanZapata/android-iconify), if you are android developer. you don't need to put a bunch of image icons in your drawable folder. You can use fontAwesome too which is already in the library list, thats why I am trying to add one more Icons Library called IonIcons. I made this as simple as possible. All you need to do is to find the icons you want and click it to copy it with three options:
- dash separator (for XML IconTextView)
- underscore separator for JAVA object
- HTML tag
- and Android-Iconify it self


Kapeli:
Failed validation: 
```
Empty name found for row id "36" at "IonIcons.docset"
```

That means that at row 36 in docSet.dsidx there's an entry without a name.

Also in docset.json please set the docset name to just "IonIcons".

bbcdanang:
some icons has no tag that's why it is empty, alright I'll delete it

bbcdanang:
I already make the changes you ask. is there something else to do?






On Sat, Oct 15, 2016 at 5:31 AM +0700, "Danang Widiantoro" <bbc_danang@fisip.net> wrote:










some icons has no tag that's why it is empty, alright I'll delete it

On Oct 15, 2016, at 05:29, Bogdan Popescu <notifications@github.com> wrote:


Failed validation: 

Empty name found for row id "36" at "IonIcons.docset"


That means that at row 36 in docSet.dsidx there's an entry without a name.

Also in docset.json please set the docset name to just "IonIcons".

PR #1142, opened 2016-10-18
update all version of ffmpeg/libav with guides

steve3d:
here is the updated docsets with all guides included.


klausbadelt:
Sorry but now some guides are listed twice (ffmpeg Documentation, ffprobe Documentation etc). Probably you're including the FFmpeg components ("*-all") versions?

steve3d:
you mean ffprobe and ffprobe-all are same and should not both be included?

klausbadelt:
@steve3d kind of. Compare 
* https://www.ffmpeg.org/ffmpeg.html
* https://www.ffmpeg.org/ffmpeg-all.html
* https://www.ffmpeg.org/ffplay-all.html
and you'll see the overlaps.

PR #1153, opened 2016-10-29
Update Sublime Text Plugin API docset for the stable Sublime Text 3 release

leonid-shevtsov:
Take 2 with fixed TOC. (Would be nice to have the verifier documented in the contribution guide)


Kapeli:
Thank you! 👍 

It's pretty straightforward to run the verifier yourself if you want to. See https://github.com/Kapeli/Dash-User-Contributions/blob/master/.travis.yml.

However, I don't recommend it. Some of the errors it gives are a bit obscure/specialised, and I prefer it if pull requests fail so I can see the errors and let the docset authors know what they can do to fix it in those cases.

PR #1156, opened 2016-11-02
ffmpeg 3.2 and libav 12 docset version update

steve3d:
update ffmpeg to 3.2 and libav to 12.

And I also removed lite version of guides, and fixed some section name error.

I don't know if this is a bug of Dash, if a section name contains character `/`, then the section are ignored, So I have to convert all `/` characters in section name to `|`. That however looks strange.

BTW, waiting the test result will make me like I'm waiting a  trial.  Ha ha .

### DIFF
--- a/docsets/FFmpeg/docset.json
+++ b/docsets/FFmpeg/docset.json
@@ -1,6 +1,6 @@
 {
     "name": "FFmpeg",
-    "version": "3.1.4",
+    "version": "3.2",
     "archive": "FFmpeg.tgz",
     "author": {
         "name": "Steve Yin",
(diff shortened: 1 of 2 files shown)


Kapeli:
Thank you! 👍 

> I don't know if this is a bug of Dash, if a section name contains character /, then the section are ignored, So I have to convert all / characters in section name to |. That however looks strange.

The entry name part of the anchors should be percent-encoded. I.e. you should use `%3E` instead of `>`. See https://kapeli.com/docsets#tableofcontents for more info.

steve3d:
@Kapeli , thanks for the info, should I update the section name? if not I'll use the correct way to handle the section names on next update.

Kapeli:
That's up to you :)

PR #1157, opened 2016-11-02
updated libav/ffmpeg docset with correct section names

steve3d:
finally I've made the section names correct now, sorry for another PR.


Kapeli:
Thank you! 👍

PR #1172, opened 2016-11-18
update to Google App Engine - Python

HyShai:
This is an update to the GAE Python docs - using the latest docs from [Github](https://github.com/GoogleCloudPlatform/google-cloud-python). It's a different source than the old docset - the old svn source has been deprecated.


Kapeli:
This replaces version 1.6.9.445 of the docset with version 0.21.0, so it's a downgrade?

@tzmartin what do you think?

HyShai:
The current API version is 0.21.0 - I think the old version was using different versioning. The old version was a "Preview".

HyShai:
Ah it seems like this codebase doesn't include the complete set of APIs - let me see if I can integrate the rest of the codebase.

tzmartin:
@HyShai Appears latest Python SDK version is 1.9.49, referencing [API release notes](https://cloud.google.com/appengine/docs/python/release-notes).

PR #1178, opened 2016-11-21
Add Should.js docset

mgduk:
https://shouldjs.github.io/


Kapeli:
The `tgz` should contain only the docset and nothing else.

PR #1196, opened 2016-12-02
Add Dash Docset for ECharts

Ovilia:
I generated two [ECharts](https://github.com/ecomfe/echarts) Docsets, each for both Chinese and English.

ECharts is a popular Data visualization tool, with more than 14,000 stars on GitHub. Users need this docset very much. Thanks in advance!


Kapeli:
Thank you! 👍 

Note: I'll change the name in docset.json to ECharts Chinese for the Chinese docset. That's the name that's displayed in Preferences > Downloads so it can't be the same for both docsets.

PR #1197, opened 2016-12-02
Improve HAProxy 1.7.0 docset documents

ruoshan:
Can I update docset of the same version? I improve some bit


Kapeli:
You can, but if you want the update to be distributed to users you need to add something to the version. Setting the version to `1.7.0/2` would show as `1.7.0` to users, but it will cause Dash to update the docset. See https://github.com/Kapeli/Dash-User-Contributions/wiki/Docset-Versioning-Guidelines for more info.

Note: this only applies to the main docset, not the specific_versions ones.

PR #1198, opened 2016-12-03
Improve HAProxy 1.7.0 docset documents

ruoshan:


Kapeli:
You changed the version wrong. You need to do it under the main `version` and not within `specific_versions`.

PR #1210, opened 2016-12-13
Add Odoo 10 docset (version 10.0.0.1/03ac47b)

vonpupp:
Greetings,

I am adding a new docsets for Odoo 10 (version 10.0.0.1/03ac47b) and Odoo 9 (version 9.0.0.1/d1eca51).


Kapeli:
Docsets shouldn't be versioned. Add a single "Odoo" folder where you'll have v10 and v9 as `specific_versions` in `docset.json`.

vonpupp:
In that case, how do you have multiple versions installed at the same time? This is a common practice with Odoo, have some clients on v8, some on v9 and start testing v10.

Kapeli:
Ah. I didn't notice there's an Odoo 8 docset. I'll allow it then.

Ideally there should have been only one docset called "Odoo" and the older versions would be under `specific_versions` which allows users to install them as well, all at the same time. See screenshot below.

screen shot 2016-12-14 at 15 16 26

Kapeli:
Thank you! 👍

vonpupp:
You are welcome.

I understand that versioning can be cumbersome. I am using Zeal and I don't have download options per version as you do. I guess this might have been the reason the original docstring was versioned in that way initially, probably the developer used Zeal also.

Best regards.

PR #1211, opened 2016-12-15
Update Terraform docset to 0.8.0

f440:


Kapeli:
Failed validation: there's an entry without a name at row number `252` in `docSet.dsidx`

PR #1233, opened 2017-01-10
Added Soot docset version 2.5.0

pengzhendong:
Here is the doc of Soot.
It is my first pull request in github, so if there is anything wrong, please tell me.
Thanks!


Kapeli:
Your docsets don't contain an `Info.plist` file. See https://kapeli.com/docsets#infoplist.

PR #1234, opened 2017-01-10
Added Soot docset version 2.5.0

pengzhendong:
I'm sorry for my negligence, because I added the Info.list in the wrong file.
Now it is correct, please check again.
Thanks!


Kapeli:
Thank you! 👍

PR #1246, opened 2017-01-20
Add Clang C++ API docset

samvv:
The same as the LLVM C++ API docset, only this time for clang.


Kapeli:
No symbolic links for icons please.

samvv:
Whoops, sorry, I forgot to remove them.

samvv:
Should be fixed now.

samvv:
@Kapeli Same for Clang:

 - https://www.dropbox.com/s/afyepayla1nrem7/Clang_API_4.0.0-rc1.docset.tgz?dl=0
 - https://www.dropbox.com/s/wrropauzlw5l586/Clang_API_3.9.1.docset.tgz?dl=0
 - https://dl.dropboxusercontent.com/u/23621256/Clang_API_3.8.1.docset.tgz

Kapeli:
The docsets look fine (with a minor exception: you shouldn't have versions in the docset name).

The problem is that the docsets are very large, especially the LLVM ones. I could accept the Clang docset, but without the `specific_versions`, but even that I'd rather not as it's not likely to be used by a lot of Dash users.

May I recommend distributing the docsets using [docset feeds](https://kapeli.com/docsets#dashdocsetfeed) instead?

PR #1256, opened 2017-02-01
Update Vulkan docset to version 1.0.39.

antiagainst:


antiagainst:
@Kapeli: It seems that the connection to GitHub experienced problems on Travis. Would you mind to restart the check?

docsets/Vulkan/docset.json
```diff
@@ -1,10 +1,20 @@
 {
     "name": "Vulkan",
-    "version": "1.0.38",
+    "version": "1.0.39",
     "archive": "Vulkan.tgz",
```
Kapeli: You need to add the archive for version `1.0.39`.

docsets/Vulkan/docset.json
```diff
@@ -1,10 +1,20 @@
...
-    "aliases": ["vk", "GLNext", "Graphics API"]
+    "aliases": ["vk", "GLNext", "Graphics API"],
+    "specific_versions": [
+        {
+            "version": "1.0.39",
+            "archive": "versions/1.0.39/Vulkan.tgz"
+        },
+        {
+            "version": "1.0.38",
+            "archive": "versions/1.0.38/Vulkan.tgz"
```
Kapeli: You need to add the archive for this version too. Moving the `txt` files I leave behind doesn't work.

docsets/Vulkan/docset.json
antiagainst: Sorry I don't quite understand. `docsets/Vulkan/versions/1.0.39/Vulkan.tgz` is already there.

Kapeli: That's the archive for the `specific_version`. You need to have it at `docsets/Vulkan/Vulkan.tgz` too for the main docset.

docsets/Vulkan/docset.json
antiagainst: You mean the original `Vulkan.tgz` for `1.0.38`?

Kapeli: Yes.

antiagainst: Well, it is already removed from the repo and now I'm adding it back again. Sounds a little bit strange, especially we have a new version now and the old version won't be used.

And just to clarify, in the future, say if I want to update 1.0.100, I need to check in all the tarballs for 1.0.38 way up to 1.0.100?

docsets/Vulkan/docset.json
antiagainst: Okey, I see. Two copies essentially. Will do.

docsets/Vulkan/docset.json
antiagainst: What probably beneficial is a readme/wiki page describing the procedure of updating docsets. :)

Kapeli: > And just to clarify, in the future, say if I want to update 1.0.100, I need to check in all the tarballs for 1.0.38 way up to 1.0.100?

Nope, no need for that. You just need to give the tarball once to the `specific_version`. Think of the stuff in `specific_versions` as completely separate from the main docset archive (i.e. the one at `docsets/Vulkan/Vulkan.tgz`).

antiagainst: Okey, I see now.

antiagainst:
@Kapeli: comments addressed. Let me know if further tweaks are needed. :)

Kapeli:
Thank you! 👍

PR #1259, opened 2017-02-08
Update Terraform docset to 0.8.6

f440:


Kapeli:
Failed validation:
```
Empty name found for row id "123" at "Terraform.docset"
```

PR #1261, opened 2017-02-12
Update TensorFlow to 1.0

ppwwyyxx:


Kapeli:
You need to add an entry for the older version to `specific_versions` (like here: https://github.com/Kapeli/Dash-User-Contributions/blob/master/Sample_Docset/docset.json#L14-L26).

Kapeli:
Thank you! 👍

PR #1263, opened 2017-02-13
add CUDA docset

blahgeek:


Kapeli:
Failed validation:
```
Found entry with new line in its name at row id "0" at "CUDA.docset". Complete entry name is: "4.1. 
                              Remote Profiling With Visual Profiler
                              "
Found entry with new line in its name at row id "0" at "CUDA.docset". Complete entry name is: "4.2. 
                              Remote Profiling With nvprof
                              "
```

This means that you have entries that contain new lines in your docset index. The row ids are wrong because they're all NULL, as well.

PR #1272, opened 2017-02-19
Corrected icon placement for Boto3

angrychimp:
Thanks again for the correction. I was apparently looking at the wrong documentation for user-submitted docsets :P


Kapeli:
Thank you! 👍

Issue #1160, opened 2016-11-03
Request: Realm (for Android)


4goodapp:
https://realm.io/docs/java/latest/
https://realm.io/docs/java/latest/api/
https://realm.io/

PR #1275, opened 2017-02-25
+ add Realm Android (Java)

blivme84:
I added Realm Java Doc to Dash.

https://github.com/Kapeli/Dash-User-Contributions/issues/1160


Kapeli:
Please use "Realm Java" for the docset name inside Info.plist and docset.json, and "Realm_Java" for the folder within this repo.

The author in docset.json should be you. It's the author of the docset, not of the docs.

Also add RealmJava.tgz at this path: `docsets/Realm_Java/RealmJava.tgz`. It's not enough to add it as a specific version.

PR #1276, opened 2017-02-25
added Linux docset

Yanpas:


Kapeli:
Please rename this to "Linux Man Pages" (in Info.plist and docset.json) and "Linux_Man_Pages" (folder name within this repo).

Please remove the `specific_versions` support. It doesn't make sense to have it for this docset.

Also hide the version from users as it doesn't make sense to display it for this docset. You can do this by starting the version with a `/` in docset.json (e.g. `/1.0.0`).

Yanpas:
@Kapeli 
1. Which strings in Plist should I rename? Renaming `DocSetPlatformFamily` seems to be the bad idea. I suggest the following:
```
<key>CFBundleIdentifier</key>
	<string>Linux</string>
	<key>CFBundleName</key>
	<string>Linux Man Pages</string>
	<string>linux</string>
```
Should I keep the folder with versions? Should I replace symbolic link in version folder to the copy of tgz arch.?

Kapeli:
1. Just the name.
2. No, delete it.

PR #1281, opened 2017-02-26
Add docset for SPIR-V

antiagainst:
https://www.khronos.org/registry/spir-v/

https://www.khronos.org/registry/spir-v/specs/1.1/SPIRV.html


Kapeli:
Thank you! 👍

PR #1288, opened 2017-02-27
Update Vulkan docset to 1.0.42

antiagainst:


docsets/Vulkan/docset.json
```diff
@@ -9,6 +9,10 @@
     "aliases": ["vk", "GLNext", "Graphics API"],
     "specific_versions": [
         {
+            "version": "1.0.42",
+            "archive": "versions/1.0.41/Vulkan.tgz"
```
Kapeli: This path is wrong. Should be `1.0.42` instead of `1.0.41`.

antiagainst: Ah, copy paste problem. Fixed

Kapeli:
Thank you! 👍

PR #1292, opened 2017-03-03
+ add Realm 3.0.0

blivme84:
Add Realm 3.0.0 document


Kapeli:
You need to update the `docset.json` too. If you want to add specific versions support follow this: https://github.com/Kapeli/Dash-User-Contributions/blob/master/Sample_Docset/docset.json#L14-L25. Please note that moving the `.tgz.txt` file around doesn't work, you need to add actual docset archives.

PR #1303, opened 2017-03-10
Update Pillow docset for Pillow v4.

153957:
Keep docs for version 3.3.0 around.


docsets/Pillow/docset.json
```diff
@@ -11,6 +11,10 @@
         {
             "version": "2.8.2",
             "archive": "versions/2.8.2/Pillow.tgz",
+        },
+        {
+            "version": "3.3.0",
```
Kapeli: Newer versions need to appear before older ones (i.e. `3.3.0` should be before `2.8.2`).

153957: Ok, done!

Kapeli:
Thank you! 👍

PR #1307, opened 2017-03-14
Add a docset for Presto, version 0.168

knu:


docsets/Presto/docset.json
```diff
@@ -0,0 +1,9 @@
+{
+    "name": "Presto",
+    "version": "0.168",
+    "archive": "Presto.tgz",
```
Kapeli: You need to add the docset at this path, not at `versions/0.168/Presto.tgz`.

knu: I see, fixed!

Issue #1308, opened 2017-03-14
Latest update of terraform docset has broken static assets


glenjamin:
See screenshot for what it looks like

<img width="1436" alt="screen shot 2017-03-14 at 13 17 56" src="https://cloud.githubusercontent.com/assets/151272/23902325/a74bddb2-08b8-11e7-99d9-4b62ade4644d.png">

Kapeli:
@f440 Can you look into this please?

f440:
Sorry, It's my mistake. Please wait for merging with #1309.

PR #1309, opened 2017-03-14
Fix broken link in Terraform docset (#1308)

f440:
#1306 cause design problem. This PR will fix it.


Kapeli:
Failed validation:

```
Checking Terraform/Terraform.tgz...
Empty name found for row id "130" at "Terraform.docset"
Checking Terraform/versions/0.8.8-1/Terraform.tgz...
Empty name found for row id "130" at "Terraform.docset"
```

This means that at row `130` in docSet.dsidx there's an entry with an empty name.

Kapeli:
Thank you! 👍

PR #1315, opened 2017-03-18
Add Clang C++ API docset 3.7.1

antiagainst:
This is the docset for Clang C++ API. It is similar to the current top of the tree C++ API at  https://clang.llvm.org/doxygen/, but for version 3.7.1. I currently need 3.7.1 to do my work. May upload updates for later versions later.

The generated `Clang.tgz` is fairly large (640MB+), so I put it in Dropbox:
https://www.dropbox.com/s/p1sg11kh7u22kq1/ClangAPI.tgz?dl=0


Kapeli:
I'm sorry but I can't accept docsets this large at this time (out of disk space).

PR #1317, opened 2017-03-18
Update to React Native 0.42.0, Draft JS v0.10.0, Redux v3.6.0-du.1

epitaphmike:
Redux v3.6.0-du.1 (doc update only)


docsets/DraftJS/docset.json
```diff
@@ -1,11 +1,27 @@
...
+        "Editor"
+      ],
+    "specific_versions": [
+        {
+          "version": "0.10.0",
+          "archive": "versions/0.10.0/DraftJS.tgz"
+        },
+        {
+          "version": "0.1.0",
+          "archive": "versions/0.1.0/DraftJS.tgz"
```
Kapeli: This archive doesn't exist.

epitaphmike: Archive 0.1.0? If you pull down the repo as is right now, there is a directory for versions/0.1.0
I am not sure what happened along the way but there was an addition of version 0.6.0/... and changes the doc set name to Draft.js.tgz from DraftJS.tgz. Do you want me to add the version/0.6.0... archive?

Kapeli: It looks like specific versions were removed in this commit: https://github.com/Kapeli/Dash-User-Contributions/commit/2d39e33a21e0d50af403c00e6725c52f23d36bd7.

My recommendation would be to just remove the specific version entry for `0.1.0` from docset.json and just have one for `0.10.0`. I can make this change if it's ok.

epitaphmike:
Sounds good.

I also sent you a link on twitter about upstream fetching for contributors.

This may be something you'd like me to try with you so we can stop deleting
and re-forking the repo on changes.

https://gist.github.com/CristinaSolana/1885435

Thanks.

Kapeli:
I don't think you should ever fetch upstream from this repo, as I'm rewriting the git history whenever I "process" a docset archive and remove it from the repo.

Kapeli:
Thank you! 👍

Issue #1326, opened 2017-03-20
Update Tensorflow docset


slcott:
https://github.com/ppwwyyxx/dash-docset-tensorflow

is newer than 

https://github.com/Kapeli/Dash-User-Contributions/tree/master/docsets/TensorFlow

Kapeli:
@ppwwyyxx Can you update when you get a chance?

ppwwyyxx:
Of course! Will update soon.

PR #1327, opened 2017-03-20
TensorFlow 1.0.1

ppwwyyxx:
#1326


Kapeli:
Thank you! 👍 

Note: newer versions should be at the top under `specific_versions`. I'll fix it.

ppwwyyxx:
btw, am I supposed to include all the old versions when creating a PR?
Since the binaries are not included upstream, it's a bit annoying to include them everytime (or maybe I'm not using git correctly)

Kapeli:
There's no need to do that. Only add docsets for new versions or for versions you actually want to change.

PR #1329, opened 2017-03-21
Added Entitas 0.40.0

sschmid:


sschmid:
Not sure why the old commits are contained in this PR. Actually just pushed `Added Entitas 0.40.0`

Kapeli:
Thank you! 👍 

> Not sure why the old commits are contained in this PR. Actually just pushed Added Entitas 0.40.0

This happens because I'm rewriting git history to remove the tgz's from it (to keep the repo size small). Have you pulled changes from the repo into your fork? If yes, try not to do that again. If not, then I'm not sure, deleting the fork and reforking seems to work.

sschmid:
Great, thanks for the info. Yes, I updated my fork first. Will not do this again ;)

PR #1333, opened 2017-03-23
add IntellJ IDEA CheatSheet (just a part of key map for now,I will complete it later)

rieonke:
Thanks~ I love dash


Kapeli:
Cheat sheets need to be added here: https://github.com/Kapeli/cheatsheets. This repo is for docsets. Sorry!

PR #1344, opened 2017-03-29
Added BrockAPI 

JustinCSafa:
Added the BrockAPI to make it a lot easier to search.


Kapeli:
Please don't contribute the actual docset. Only an archive of it.

PR #1347, opened 2017-03-30
pull swoole-docset.1.0.0

halfstring:
pull request: swoole-docset 1.0.0


Kapeli:
Please change the name of the docset everywhere to "Swoole Chinese". Non-english docs need to specify their language to avoid confusion.

Also you need to change the `icon.png` to be `16x16` in size and `icon@2x.png` to be `32x32`.

Also the docset author in `docset.json` needs to be you, not "swoole team"

PR #1363, opened 2017-04-11
Update Chainer version to v1.23.0

mitmul:
The newest Chainer v1.23.0 has been released today.
This PR updates the documents also.


Kapeli:
Thank you! 👍 

One note: versions under `specific_versions` need to be ordered by newest-first. I'll change this.

PR #1368, opened 2017-04-13
Force update HAProxy lua docset

ruoshan:
this docset is still unstable, it's v1.0, but the content might be
changed from upstream.


Kapeli:
Thank you! 👍

PR #1384, opened 2017-04-25
glibc 2.25

davidwed:


docsets/glibc/docset.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "glibc",
-    "version": "2.24",
+    "version": "2.25“,
```
Kapeli: `“` isn't a regular quote symbol. It's smart quote or something like that. Notice the difference between `“` and `"`

Kapeli:
Thank you! 👍

PR #1387, opened 2017-04-27
Update LLVM Core

Naville:
The zip is  128.3MB when compressed using 7z, which I found reasonable compared to other 700mb+ LLVM PRs 
Please refer to https://mega.nz/#!DhNSDbCC!26ajVUdxkj3EniYCqSvaZIHZwQLpy_4MOTIOYQ44jA8

### DIFF
--- a/docsets/LLVM/README.md
+++ b/docsets/LLVM/README.md
@@ -1,4 +1,15 @@
-LLVM Dash Docset
+Original LLVM-Dash was last updated on 9 Sep 2015, more than two years ago.
+Unlike original version , this one was built according to the official guide with some modifications.
+See <http://mayuyu.io/llvm/2017/04/06/BuildingLLVMDoc.html>
+
+
+Below was the original README
+
+
+Naville Zhang
+
+
+## LLVM Dash Docset
 =======================
 
 - __Docset Description__:
--- a/docsets/LLVM/docset.json
+++ b/docsets/LLVM/docset.json
@@ -1,10 +1,10 @@
 {
     "name": "LLVM",
-    "version": "3.7",
-    "archive": "LLVM.tgz",
+    "version": "5.0SVN301109",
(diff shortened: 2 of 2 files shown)


Naville:
I've already read about the disk space issues in other rejected PRs. (Which I find unreasonable considering Dash is a paid app and it's not cheap)

As such I've compress this in 7z format, if you could unpack it client size you shouldn't have issues with storage space

Naville:
Or , may i suggest add more options to dash so we can:
- Host JSON on github or your server 
- put the actual files somewhere else.

Mega's SDK is open source so I assume there shouldn't be too much trouble implementing this

Naville:
Note 7Z is LGPL Licensed so you should have no issue dynamically linking it, if that's a issue

Kapeli:
Dash currently doesn't support 7Z archived docsets.

> Or , may i suggest add more options to dash so we can:
> 
> Host JSON on github or your server
> put the actual files somewhere else.
> Mega's SDK is open source so I assume there shouldn't be too much trouble implementing this

See https://kapeli.com/docsets#dashdocsetfeed.

Naville:
tgz results in over 600MB archive last time I tried

Naville:
https://mega.nz/#!W1UUyDoR!WX-imkB8i8H48SB2mwtodAx_KHo9d5_J8Xn0fHyW2L0 Improved to be under 300mb

Naville:
> Docset feeds should only be used for docsets which only you, your team or very few users find useful.

Kapeli:
The docset fails validation. It has entries in docSet.dsidx that point to paths that don't exist.

Run this command in the root folder of this repo to run validation locally:

```
wget http://kapeli.com/feeds/zzz/docsetcontrib.tgz && tar -xzf docsetcontrib.tgz && ./docsetcontrib --verify
```

PR #1388, opened 2017-04-28
Add TensorFlow API 1.1 docset

fossil12:
I used @ppwwyyxx's script to create the docset of TensorFlow's most recent release. Thus I though I can also contribute it here.


Kapeli:
Failed validation. A lot of entries contain newlines in their names, which isn't supported.

fossil12:
@Kapeli Thank you for the feedback. I opened a new PR with the fixed docset.

PR #1390, opened 2017-05-01
Add TensorFlow API 1.1 docset

fossil12:
Follow up on PR #1388. I fixed the newlines problem.


Kapeli:
Moving the `tgz.txt` in `specific_versions` isn't supported, you need to add the actual `tgz` for the `specific_version`. I'll fix it this time.

fossil12:
@Kapeli Ok, good to know. Thanks.

PR #1414, opened 2017-05-17
Update FileMaker docset to version 16.0.1

jpscharf:


Kapeli:
You need to add the docset archive at this path as well: `docsets/FileMaker/FileMaker.tgz`. It's not enough to add it in the `versions` folder.

jpscharf:
@Kapeli I added the updated archive to the root of the FileMaker docset.

Kapeli:
Thank you! 👍

PR #1422, opened 2017-05-22
Update Atom docset to 1.17.0

bkbooth:
I've switched back to my generation script.

I've improved it quite a lot so I should be able to keep on top of pushing new versions pretty quickly now.


Kapeli:
The `icon.png` needs to be `16x16` and you also need to have an `icon@2x.png` file that's `32x32`. It's ok if I fix this?

bkbooth:
Yep, sorry I missed this. Happy for you to fix it if you want to do it.

Kapeli:
Thank you! 👍

PR #1432, opened 2017-06-02
Ramda 0.24.1

alexbepple:


Kapeli:
Thank you! 👍 

Note: the specific versions need to be ordered by newest-first. I'll fix it.

alexbepple:
Thanks for fixing the ordering. However, as it's important, it would be great if the build could check for that. Is there a case where it is not possible to verify the order automatically?

Kapeli:
I couldn't find a way to handle sorting of versions in all of its shapes and forms (e.g. versions that contain dates, or beta/alpha info). If I'd have a reliable way to do that, I'd be able to just sort the versions on Dash's end and have no requirement for the json.

I know it's doable, but it's that kind of minor issue that's not worth the time investment.

I check and fix the versions myself currently. Nothing happens if the order is different, it's just displayed as-is in Dash. I like to have them latest-first, so that they're all consistent in Dash.

alexbepple:
Yeah, that's what I thought. ;) Thanks.

PR #1437, opened 2017-06-06
Add an new docset ICU 59.1

fantasticfears:


docsets/ICU/docset.json
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "ICU",
+    "version": "59.1/20170606",
+    "archive": "org.icu-project.icu4c.tgz",
```
Kapeli: This file doesn't exist. You added "ICU.tgz", not "org.icu-project.icu4c.tgz".

Kapeli:
The docset name (in Info.plist) shouldn't contain a version number.

Kapeli:
There are entries that point to files that don't exist. For example, "icu::AlphabeticIndex".

PR #1442, opened 2017-06-08
Add Docset for AWS CloudFormation Template Reference

pdhodgkinson:
Docset for AWS CloudFormation Template Reference as found here:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-reference.html


Kapeli:
Thank you! 👍

PR #1460, opened 2017-06-28
feature: add knex docset

michaelgmcd:
Let me know if there are any other things needed for this to be merged!


docsets/Knex/docset.json
```diff
@@ -0,0 +1,23 @@
...
+    "knexjs",
+    "knex.js",
+    "bookshelf",
+    "bookshelfjs",
+    "bookshelf.js"
+  ],
+  "specific_versions": [
+    {
+      "version": "1.0.0",
+      "archive": "versions/1.0.0/Sample.tgz"
```
Kapeli: This should be changed to `versions/1.0.0/Knex.tgz`

docsets/Knex/docset.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "Sample",
```
Kapeli: Name should be "Knex"

Kapeli:
The docset archive should only contain the docset and nothing else. Currently, the archive contains some folders which contain the docset.

PR #1469, opened 2017-07-04
Add Manual for GNU Emacs Common Lisp Emulation

pekingduck:
This manual is not included in the main Emacs docset.

Source: https://www.gnu.org/software/emacs/manual/html_node/cl/index.html


Kapeli:
Thank you! 👍

PR #1477, opened 2017-07-10
Update Requests docset to 2.18.1

medeirosthiago:
Hey guys,
This is my first PR.
I added a `versions` folder with the old docset and the new docset that I updated.
If I made some mistake I'm open to do it right! 😬


Kapeli:
Looks good. Thank you! 👍

PR #1502, opened 2017-08-02
Add Riot.js docset

Vic-Dev:


Kapeli:
Please delete the `versions` folder if you're not using it.

The docset index contains a lot of invalid entries, like `404: Page not found`, `Opening Plunker...`. Also it contains a lot of entries in different languages. Please stick to one language only within a docset. If the language you choose is not English, you need to clearly mark it in the docset name.

PR #1519, opened 2017-08-16
Add TensorFlow 1.3

ppwwyyxx:


Kapeli:
Failed validation. A lot of entries in the docSet.dsidx point to files that don't exist. See https://travis-ci.org/Kapeli/Dash-User-Contributions/builds/265318267 for the log.

ppwwyyxx:
The first error in the log shows this file doesn't exist, but it does. What's going on? 
```
╰─$wget https://github.com/ppwwyyxx/Dash-User-Contributions/raw/ddd9070cfd68804b9644fdcd7bc64c17685b16b3/docsets/TensorFlow/TensorFlow.tgz
╰─$tar xzf TensorFlow.tgz
╰─$find | grep versions/master/api_docs/python/tf/Summary/Audio.html
./TensorFlow.docset/Contents/Resources/Documents/versions/master/api_docs/python/tf/Summary/Audio.html
```

Kapeli:
Your archive contains both "versions/master/api_docs/python/tf/Summary/Audio.html" and "versions/master/api_docs/python/tf/summary/audio.html". Most Macs are case-insensitive, so you shouldn't have that.

ppwwyyxx:
Ah I see. Thanks!
These are indeed two different pages:
https://www.tensorflow.org/versions/master/api_docs/python/tf/Summary/Audio
https://www.tensorflow.org/versions/master/api_docs/python/tf/summary/audio

I'll see how I can work around it..

Issue #1524, opened 2017-08-18
Powershell modules


lucasg:
Not an issue *per se*, more asking for some clarifications.

I've written a docset generator for the ["Powershell modules" API browser](https://docs.microsoft.com/en-us/teamblog/announcing-unified-powershell-experience) and I was thinking of contributing back if your're interested.

* Powershell currently support several API versions (`3`,`4`,`5.0`,`5.1` & `6`) and the docs are versioned also. What is preference for versioning docset ? Do I generate a `Powershell3.tgz`, `Powershell4.tgz`, etc. like for Python2/Python3 ? Or do every docset have the same name ?
* Microsoft keep on using the word `Cmdlets` to designate Powershell commands/functions (maybe to differentiate them from traditionnal batch commands). Is it possible to add an [entry type]( https://kapeli.com/docsets#supportedentrytypes) for it ? Otherwise, I will use the `Command` entry type.
* I used `posh` as the `CFBundleIdentifier` since `Powershell` is often abbreviated as such ( *a la* `csh`, `zsh`, etc.). Is that okay ?

repo : https://github.com/lucasg/powershell-docset
docsets generated : https://github.com/lucasg/powershell-docset/releases

Kapeli:
> Powershell currently support several API versions (3,4,5.0,5.1 & 6) and the docs are versioned also. What is preference for versioning docset ? Do I generate a Powershell3.tgz, Powershell4.tgz, etc. like for Python2/Python3 ? Or do every docset have the same name ?

I'd say use [`specific_versions`](https://github.com/Kapeli/Dash-User-Contributions/blob/master/Sample_Docset/docset.json#L14-L26). You can also make separate docsets for each major version if you must though.

> Microsoft keep on using the word Cmdlets to designate Powershell commands/functions (maybe to differentiate them from traditionnal batch commands). Is it possible to add an entry type for it ? Otherwise, I will use the Command entry type.

Best if you use the Command type. I can definitely add the Cmdlet type if you want me to, but I don't think it's a good idea to add a type that's specific only to one docset.

> I used posh as the CFBundleIdentifier since Powershell is often abbreviated as such ( a la csh, zsh, etc.). Is that okay ?

Yep.

PR #1525, opened 2017-08-19
Powershell modules docset

lucasg:
Referencing : https://github.com/Kapeli/Dash-User-Contributions/issues/1524


Kapeli:
`isDashDocset` needs to be `true`, not `false` (in your docsets' Info.plist).

lucasg:
Okay, I've updated to `true`. How is this key used ( I'm pretty sure `Velocity` or `Zeal` don't parse it) ?

Kapeli:
Thank you! 👍 

One note: the `specific_versions` need to be ordered by newest-first. I'll fix that.

> Okay, I've updated to true. How is this key used ( I'm pretty sure Velocity or Zeal don't parse it) ?

It's used to differentiate between Apple docsets and Dash docsets as they have different database formats. See https://github.com/Kapeli/Dash-iOS/search?q=isdashdocset.

PR #1532, opened 2017-08-21
Adding Salesforce Lightning docs

mattgoldspink:
This is an initial version of adding docs for Salesforce's Lightning platform. Original docs are here: https://developer.salesforce.com/docs/atlas.en-us.lightning.meta/lightning/intro_framework.htm

Let me know if I've missed anything for this to be contributed.


docsets/Salesforce_Lightning/docset.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "Salesforce Lightning",
+    "version": "40.0",
+    "archive": "Salesforce_Lightning.tgz",
```
Kapeli: There's no `Salesforce_Lightning.tgz` in the main folder. You need to add it. Also remove `Sample_Docset.tgz`.

Kapeli:
All good otherwise.

docsets/Salesforce_Lightning/docset.json
mattgoldspink: Good catch! Should be fixed now

Kapeli:
Thank you! 👍

PR #1543, opened 2017-09-02
Update Atom docset to 1.19.5

bkbooth:
There seems to have been a lot of patch versions with Atom v1.19, would you prefer if I only send through minor versions? It's only a few mins for me to generate them now, so I'm not too fussed either way.


Kapeli:
Thank you! 👍 

Send as many versions as you want. It's no work at all for me to accept them.

bkbooth:
Ok thanks. Good to know.

PR #1546, opened 2017-09-05
Zsh 5.4.1 (new)

zmwangx:
Unfortunately I can only find a 16x16 icon (the favicon of zsh.sourceforge.net). `icon@2x.png` in this repo is upscaled, but I did not put the upscaled version into the actual docset.


Kapeli:
Thank you! 👍 

I'll remove the icon completely. The upscaled version isn't ok and there must be a 16x16 and a 32x32 version.

PR #1548, opened 2017-09-06
Add Kubernetes API docset

ruoshan:
there is an existing Kubernetes docset that's quite outdated. this
docset is a light one, with only API reference (up to date).


Kapeli:
I don't think we should have 2 docsets for the same thing. Ideally the existing docset should be updated or replaced.

// cc @nextrevision

ruoshan:
Yeah~ agree.

nextrevision:
👍 will look into it today

nextrevision:
I'm fine with replacing the existing kubernetes docset with this. Feel free to replace mine.

Kapeli:
@ruoshan Please make a new pull request that replaces @nextrevision's docset with yours.

ruoshan:
Okay, will do. Thanks you two.

PR #1552, opened 2017-09-10
Updating Georg Barikin's docset to the newest version of Chrome's API…

heyoni:
… while also adding user guides to the docset


Kapeli:
Thank you! 👍 

// cc @gebrkn

PR #1580, opened 2017-09-22
Changed path of index.html in Info.plist

almai:
Changed path of index.html in Info.plist and created archive for docset again.


Kapeli:
Thank you! 👍

PR #1585, opened 2017-09-26
Update MXNet docsets version to 0.11.0 and add 'versions' directory

yafz:


Kapeli:
The format for `specific_versions` in `docset.json` is wrong. You also need to update the main `MXNet.tgz`.

PR #1587, opened 2017-09-27
Updated Terraform Docset to v0.10.6

rolandjohann:
Terraform doc management has been changed since v0.10.0 and required several changes to use the existing code and generate the dash docset.

In addition, the generator script contains a fix for separating terraform data and resources at the SQlite index. This requires that the old docs should be regenerated - is it possible to just replace the old docsets?

### DIFF
--- a/docsets/Terraform/README.deprecated.md
+++ b/docsets/Terraform/README.deprecated.md
@@ -1,0 +1,17 @@
+Terraform Docset
+=======================
+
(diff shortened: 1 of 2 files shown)


Kapeli:
You need to update the `docset.json`. You can replace older versions of docsets if you want.

Make sure to also update the main `Terraform.tgz`.

PR #1589, opened 2017-09-28
Initial load of LiveCode docset

jameshale:


Kapeli:
Failed validation:

```
Found directory in main index: Entry with row id "1991" and path "" at "Livecode.docset"
Empty type found for row id "2237" at "Livecode.docset"
Empty type found for row id "4692" at "Livecode.docset"
Empty type found for row id "5348" at "Livecode.docset"
```

PR #1590, opened 2017-10-01
Updated Terraform Docset to v0.10.6

rolandjohann:
Terraform doc management has been changed since v0.10.0 and required several changes to use the existing code and generate the dash docset.

In addition, the generator script contains a fix for separating terraform data and resources at the SQlite index.
Previous docset versions will be fixed in a separate pull request.


Kapeli:
Thank you! 👍

PR #1593, opened 2017-10-02
esoftplay documentation

esoftplay:
please approve this one


esoftplay:
Please help me.. I don't understand, what's wrong with my commit. It looks everything is fine

Kapeli:
Failed validation:

```
File not found at "_CLASS.html" for entry with row id "107" of "Esoftplay.docset"
File not found at "_FUNC.html" for entry with row id "108" of "Esoftplay.docset"
File not found at "_LIB.html" for entry with row id "109" of "Esoftplay.docset"
```

This means that you have entries in your `docSet.dsidx` that point to files that don't exist (case-sensitive). For example, there's no `_CLASS.html` file in your Documents folder. There's only `_class.html`.

Please don't email me regarding pull requests. I'll reply here, there's no need to also email me.

esoftplay:
I just find out that the default format for HFS+ in macos is case insensitive. So must use linux to recreate my docsets

Kapeli:
Thank you! 👍

PR #1595, opened 2017-10-03
Adjusted initial load for LiveCode docs

jameshale:
First attempt had errors in the SQLITE file that revealed errors in the actual documentation.
These have now been accounted for.
documentation as of LC 9dp9


jameshale:
aurgh! index file mangled.

PR #1604, opened 2017-10-07
livecode v1.01

jameshale:


docsets/LiveCode/docset.json
```diff
@@ -12,6 +12,10 @@
         {
             "version": "1.0.0",
             "archive": "versions/1.0.0/LiveCode.tgz"
-            }
+            },
+         {
+             "version": "1.0.1",
```
Kapeli: Newer versions need to be at the top of this list, not bottom.

jameshale: Just to check. This is the only place where a new version is specified (apart from its actual presence)?

Kapeli: Yep

Kapeli:
Thank you! 👍

PR #1611, opened 2017-10-17
Update SpigotMC docset to 1.12.2

Synapz1:
When I cloned, the past version was named SpigotMC.tgz.txt. I'm not sure if this was automatic so I was confused if it should be renamed back to SpigotMC.tgz when placed in versions for archiving. I left it as SpigotMC.tgz.txt and just moved it to version/1.10.2 and added the new updated version as SpigotMC.tgz in the root directory. I think this is right, but I updated because Spigot-1.12.2 has many updates that are not shown in Spigot-1.10.2. Hopefully there are no issues with the name of the file. Thanks


Kapeli:
Your docset is not indexed and does not contain an Info.plist file.

You need to also add `versions/1.10.2/SpigotMC.tgz`. The `txt` files can be deleted.

PR #1616, opened 2017-10-18
Updates path.py to version 10.4

rspeed:


Kapeli:
Moving the `txt` file to `versions` doesn't work. You need to add the actual `tgz`.

PR #1632, opened 2017-10-31
Added Ramda docset

Akronix:


Kapeli:
Why did you replace the icon with a red block?

Kapeli:
@alexbepple Could you update the Ramda docset please?

Akronix:
OMG! I didn't know that there was already a docset for Ramda :flushed: 
It definitely doesn't appear in [Zeal](https://github.com/zealdocs/zeal/), I'll investigate why.

PR #1642, opened 2017-11-05
Added Chainer 3.0.0 docs

oroszgy:


Kapeli:
You need to update the `docset.json` and the main archive as well.

oroszgy:
Uh, I missed that, sending another PR with the fixed updates.

PR #1644, opened 2017-11-06
Add Ramda Chinese

8881:


Kapeli:
Failed validation: a lot of the entries in `docSet.dsidx` contain entries with newlines in their names. See https://travis-ci.org/Kapeli/Dash-User-Contributions/builds/297830586 for full log.

PR #1654, opened 2017-11-15
update Hapi docset to 17.0

sundeer:
wanting to start using versions if that's ok


Kapeli:
Your docset doesn't seem to be properly indexed. It's missing Methods, Properties and Constructors from its index, which the old docset had.

// cc @z0mt3c

sundeer:
It matches what I'm seeing on the Hapi api docs page. Many things have changed going from v13 to v17. How 'bout if I submit a new pr with all the intermediate versions from 13-17 if I'm able?

Kapeli:
What I said isn't regarding the doc pages themselves, but the docset index.

See screenshot below. Your docset is the one at the top. It has only 5 indexed entries.

screen shot 2017-11-15 at 21 08 32

z0mt3c:
Stop using dash some time ago... Source is located here https://github.com/z0mt3c/hapi.docset - Its just parsing the (excellent) API.md and build an index. Nothing special just quick and easy. Probably some of the regular expression have to be changed.

sundeer:
ok, I see now. Hapi doc format changed with v17, I'll look at it.
Thanks @z0mt3c

PR #1659, opened 2017-11-16
Update readme with correct project name

lb-:
Project name was set to 'Pillow' but should be 'Wagtail' in readme.


Kapeli:
Thank you! 👍

PR #1664, opened 2017-11-17
Clarification about docset archive locations (Wagtail)

lb-:
Also moved the icon note to its own heading as it is not really part of the normal instructions


Kapeli:
Thank you! 👍

PR #1669, opened 2017-11-29
add ffmpeg 3.4 docs and bump libav  to 12.2

steve3d:
hope this passes

### DIFF
--- a/docsets/FFmpeg/docset.json
+++ b/docsets/FFmpeg/docset.json
@@ -1,6 +1,6 @@
 {
     "name": "FFmpeg",
-    "version": "3.3.2",
+    "version": "3.4",
     "archive": "FFmpeg.tgz",
     "author": {
         "name": "Steve Yin",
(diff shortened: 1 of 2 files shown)


Kapeli:
Failed validation:

```
File not found at "api/structFragment.html" for entry with row id "47238" of "FFmpeg.docset"
```

steve3d:
looks like the docs need some manual adjustment. hope this time passes.

Kapeli:
Same error. Please note that the paths are case-sensitive.

steve3d:
I know that, but in ffmpeg there is this `struct fragment` and `struct Fragment`, they both created a api page with `structfragment.html` and `structFragment.html`, and I'm pretty sure , your test are based on case-sensitive but your mac partition is not, so, if the `structfragment.html` extracted after `structFragment.html`, then on a case-insensitive mac partition, the actual file becomes `structfragment.html`. that's why this test will fail. 

on actual mac machine, almost everyone use a case-insensitive partition,  these two files both will work, but there must be one pointing to the wrong api document.

I'll improve the build script to manually change the file name.

steve3d:
finally this problem is permanently fixed in my generator scripts.

Kapeli:
Thank you! 👍 

> on actual mac machine, almost everyone use a case-insensitive partition, these two files both will work, but there must be one pointing to the wrong api document.

Yep, but I check this because docsets are used on case-sensitive platforms too, like Linux (through Zeal) and iOS.

PR #1677, opened 2017-12-06
Updated Discord4J to support all version since 2.7

jammehcow:


docsets/Discord4J/docset.json
```diff
@@ -1,14 +1,42 @@
 {
     "name": "Discord4J",
-    "version": "2.7.0",
+    "version": "2.9.1",
     "archive": "Discord4J.tgz",
```
Kapeli: You need to update the main archive as well.

jammehcow: Fixed in #3eb2f7a

Kapeli:
Thank you! 👍

PR #1698, opened 2017-12-27
Updated capture for widget and library summaries

jameshale:
Seems a documentation convention has changed as noted in README


Kapeli:
Thank you! 👍

PR #1708, opened 2018-01-04
Update docset.json

kogent:
Update "version" to latest: 0.11.1

Unless there is a reason to have the default version not be the latest?


Kapeli:
The `Terraform.tgz` is for version `0.10.8`. You need to also update the `tgz` if you want to update the version.

PR #1721, opened 2018-01-15
Updated ns-3 to version 3.27.

tnull:


docsets/ns-3/docset.json
```diff
@@ -12,6 +12,10 @@
 		{ 
 			"version": "3.25",
 			"archive": "versions/3.25/ns-3.tgz"
+		},	
+		{ 
+			"version": "3.26",
+			"archive": "versions/3.26/ns-3.tgz"
```
Kapeli: Newer versions should be at the top of this list.

Also, you need to add the actual `tgz` for this version. Moving the `txt` file doesn't work.

tnull:
Should be better now?

Kapeli:
Thank you! 👍

tnull:
I found a little bug in my docset (it used the old Info.plist and therefore referenced the old fallback url). I updated my branch to fix this.

PR #1725, opened 2018-01-17
Fixed the ns-3 docset, since it still seemed to be the old.

tnull:
I found a that my docset downloaded through Dash is still the old one. Maybe I messed something up and accidentally added the old instead of the new docset?! 

However, I updated my branch to fix this.


Kapeli:
Thank you! 👍

PR #1733, opened 2018-01-29
Update enzyme docset to 3.3.0

Gidgidonihah:
I've also written a proper generation script. It's now included so anybody can fairly easily update versions.


Kapeli:
Thank you! 👍

PR #1742, opened 2018-02-02
adjust for nomenclature change

jameshale:
widget libraries are now typed as "modules". Control Structures need to be capitalized.


Kapeli:
Thank you! 👍

PR #1747, opened 2018-02-08
Update TensorFlow 1.5

ppwwyyxx:


Kapeli:
The archive you add needs to contain the docset at its base. The archives you added contain the docset under the `www.tensorflow.org` folder.

ppwwyyxx:
Ah yes.  Am I supposed to open a separate PR after the fix?

Kapeli:
Yes, please.

PR #1757, opened 2018-02-10
Update WP-CLI

ptrkcsk:


ptrkcsk:
@Kapeli `docsets/WP-CLI/WP-CLi.tgz` is a copy of `docsets/WP-CLI/versions/1.5.0/WP-CLI.tgz` because v1.5.0 is the latest. Am I doing the versioned docsets correctly? Should I omit one of those archives to avoid redundancy?

Kapeli:
Thank you! 👍 

Nope, that's correct. You should always include both archives, even if they're the same.

ptrkcsk:
Great. Thanks!

PR #1765, opened 2018-02-16
Update Blender docset to 2.79

alexwidener:
This will add https://docs.blender.org/api/2.78/

I'm about to add 2.79 as well in a separate MR


alexwidener:
Ugh, I forgot that submitting any subsequent commits ties to the upstream repository. I submitted 2.78 and then submitted 2.79 - will it work fine, as it is, with 2.78 in the version directory?

Kapeli:
All good. Thank you! 👍 

Some feedback: please consider setting a different `dashIndexFilePath`. The default one (`genindex.html`) is very large and locks the UI while loading.

PR #1769, opened 2018-02-17
Adds pandas 0.22.0 documentation, with versioning to support updates

allank:
Latest pandas documentation


Kapeli:
A Pandas docset is already supported as part of the main Dash docsets (see screenshot).

Any thoughts on why your docset should be included as well?

screen shot 2018-02-17 at 14 15 02

allank:
I added it because I didn't find pandas when I tried to add it:

pandas_missing

I'm using Dash `2.2.6` Full version (activated), not sure why it doesn't show up for me when I search for it.

Kapeli:
Pandas is available in Dash starting with version 3. You need to update to Dash 3 or newer.

PR #1775, opened 2018-02-25
Update Discord4J to 2.9.3

jammehcow:


jammehcow:
It for some reason kept commits from the last PR.

Kapeli:
Thank you! 👍 

> It for some reason kept commits from the last PR.

I think that's due to the way I rewrite the git history to remove the `tgz`s (to keep the repo size small).

PR #1783, opened 2018-03-01
Add Solidity Docset 0.4.21 

oschvr:
- Tested .dsidx in DB Browser for Sqlite
- Tested .docset locally in Dash


Kapeli:
Thank you! 👍

PR #1803, opened 2018-03-16
First commit of babylonjs docset

bnolan:
Docs for babylonjs 3.1.


Kapeli:
The `icon.png` needs to be `16x16` in size and `icon@2x.png` needs to be `32x32`.

The CSS of the index page of the docset doesn't work (see screenshot).

screen shot 2018-03-17 at 14 23 14

PR #1812, opened 2018-03-21
Include docs for NetworkX 2.1

AllanLRH:
I've tried to add the updated docset as new version of the the old docset, preserving both.
It's a major update from 1.11 to 2.1, so some projects might still user the old version.

I've also updated the icon to match the graph that's greeting users when they visit the NetworkX homepage – I don't think that NetworkX have an official logo/icon.

This is my first pull request to Dash-User-Contributions, so please review with extra scrutiny ;)


docsets/NetworkX/docset.json
```diff
@@ -1,15 +1,21 @@
 {
     "name": "NetworkX",
-    "version": "1.11",
-    "archive": "NetworkX.tgz",
+    "version": "2.1",
+    "archive": "versions/2.1/NetworkX.tgz",
```
Kapeli: This is not currently supported. The archive needs to be different from the `specific_versions` ones.

docsets/NetworkX/docset.json
```diff
@@ -1,15 +1,21 @@
...
+    "aliases": [],
+
+    "specific_versions": [
+        {
+            "version": "2.1",
+            "archive": "versions/2.1/NetworkX.tgz",
+        },
+        {
+            "version": "1.11",
+            "archive": "versions/1.11/NetworkX.tgz",
```
Kapeli: You need to add the actual `tgz` at this path. Moving the `txt` file around is not supported.

docsets/NetworkX/README.md
```diff
@@ -1,13 +1,12 @@
-NetworkX Docset
-===============
+Sample Docset
```
Kapeli: You seem to have replaced the README with the sample one?

Kapeli:
The version number should not be included in the docset name in `Info.plist`.

docsets/NetworkX/docset.json
AllanLRH: Should there be a version 2.1 in both the NetoworkX-folder and in a version-folder?

docsets/NetworkX/docset.json
AllanLRH: How do I get that file?

docsets/NetworkX/docset.json
Kapeli: The archive is a duplicate, yes.

docsets/NetworkX/docset.json
Kapeli: https://kapeli.com/feeds/zzz/user_contributed/build/NetworkX/NetworkX.tgz

AllanLRH:
This is the current info.plist. Notice that the version number is only removed from the tag with _name_ in it's name.
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>dashIndexFilePath</key>
    <string>index.html</string>
    <key>CFBundleIdentifier</key>
    <string>NetworkX 2.1</string>
    <key>CFBundleName</key>
    <string>NetworkX</string>
    <key>DashDocSetFallbackURL</key>
    <string>https://networkx.github.io/documentation/stable/</string>
    <key>DashDocSetFamily</key>
    <string>python</string>
    <string>networkx 2.1</string>
    <key>isDashDocset</key>
    <true/>
    <key>isJavaScriptEnabled</key>
    <false/>
</dict>
</plist>
```

Kapeli:
Thank you! 👍

PR #1819, opened 2018-03-27
Updated AioHTTP to version 3.1.0.

AllanLRH:
Major version jump, from 2.X to 3.X, 3.0.0 released 2018-02-12.
Also updated the icon, to reflect the docs-page (the icon changed when v. 3.X were released).


Kapeli:
The docsets should not contain any version numbers (e.g. in the docset name or the Info.plist). The only place you should really have versions is `docset.json`.

Kapeli:
Thank you! 👍

PR #1865, opened 2018-04-27
Update Presto docset to 0.200

knu:


knu:
Sorry, just filed #1866!

PR #1871, opened 2018-05-02
Added a new docset for DITA

pmazaitis:
Please take a look at the DocSet for DITA (Darwin Information Typing Architecture); I _think_ I have everything in the right place, but this is my first one of these, and I'd like to make sure I've not broken anything, or if there is something I should amend or fix.

Thanks!


Kapeli:
Thank you! 👍

PR #1875, opened 2018-05-04
fix: Update docset to optimize its display

singularitti:
Before this fix, search in the docset will direct
to a page, but with navbar blocking the result.
This is caused by an option `'navbar_fixed_top': "true"`
in Sphinx Bootstrap Theme, now this bug
has been fixed in this version of docset.
The conflicts have been resolved.

Conflicts:
-	docsets/Numba/Numba.tgz
-	docsets/Numba/versions/0.38.0/Numba.tgz


Kapeli:
Thank you! 👍

PR #1881, opened 2018-05-08
openFrameworks 0.10.0 added

icq4ever:
openFrameworks 0.10.0 added


docsets/openFrameworks/docset.json
```diff
@@ -9,9 +9,13 @@
     "aliases": ["of"],
 
     "specific_versions": [
+		{
+			"version": "0.10.0",
+			"archive": "versions/0.10.0/openFrameworks.tgz"
```
Kapeli: You need to add an archive at this path as well.

docsets/openFrameworks/docset.json
```diff
@@ -9,9 +9,13 @@
...
+		},
     	{
-    		"version": "0.9.8",
-    		"archive": "versions/0.9.8/openFrameworks.tgz"
+			"version": "0.9.8",
+			"archive": "versions/0.9.8/openFrameworks.tgz"
```
Kapeli: You need to add an archive at this path as well.

docsets/openFrameworks/docset.json
icq4ever: ok. I copied to that path as well.

Kapeli:
Thank you! 👍

PR #1886, opened 2018-05-15
added a new Julia docset

cormullion:
This is a Dash docset for Julia. (The previous Julia docset was apparently made for version Julia 0.3, which was back in 2015.)

This version is for the "just about to be released in alpha" version 0.7.


Kapeli:
I'll update the main Julia docset to 1.0.0 as soon as it becomes stable.

cormullion:
It would be nice to have the current (v0.7) docs available now so that people can refer to them when updating their packages... The current version is no use at all.

Kapeli:
The Julia docs available in Dash are for version 0.6.2 (in the main docsets). I'm going to remove the user contributed Julia docset (I forgot it exists).

As far as I can tell the latest docs for Julia are 0.6.2 - https://docs.julialang.org/en/stable/. Where do you get the 0.7 docs from?

cormullion:
Oh, I must have found that earlier version, sorry... It's true that version 0.6.2 is current, but the latest version is at: https://docs.julialang.org/en/latest/ . Version 0.7 has many sweeping changes, which is why I started being interested in faster documentation, hence Dash... :)

Kapeli:
I see. I'll update to 0.7.0 ASAP.

PR #1896, opened 2018-05-22
Contributed docset for Python Materials Genomics

shyuep:
First version of contributed docset for pymatgen (Python Materials Genomics). Latest version v2018.5.22.


Kapeli:
Thank you! 👍

PR #1900, opened 2018-05-23
Octave v4.2.2 & v4.4.0

obstschale:
Update Octave docset.

New Versions:

* v4.2.2
* v4.4.0

Yes, there is no v4.3. I don't know why ¯\_(ツ)_/¯


docsets/Octave/docset.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "Octave",
-    "version": "4.2.1",
+    "version": "4.4.0",
     "archive": "Octave.tgz",
```
Kapeli: You need to update the main archive as well.

Kapeli:
Thank you! 👍

PR #1901, opened 2018-05-23
add pwntools docset and improve IntelASMx86

Kira-cxy:


Kapeli:
Please don't rename the folder of the docset you're updating/improving (leave it as `IntelASMx86`). Dash uses the folder name as the unique identifier, so changing it makes Dash treat the docset as a completely new one.

Kira-cxy:
I thought the origin name cannot describe the docset precisely, so I did the change. Wish to rename it.

Kapeli:
The folder name isn't displayed to users at all, it's just used internally by Dash as an identifier. You should rename the name in `docset.json` instead if you want.

Kapeli:
Thank you! 👍

PR #1924, opened 2018-06-09
Update FreeBSD Handbook r51776

publicarray:


Kapeli:
You need to update the version in `docset.json` as well. Otherwise users don't get the update.

publicarray:
Thanks, I changed the version to the revision. Initially I was going to track the FreeBSD version this handbook would be applied to. But I think it makes more sense to track the handbook revision since it is a living document with frequent but small updates

Kapeli:
Thank you! 👍

PR #1948, opened 2018-06-25
Carbon 1.31.0

Max13:
Make use of specific_version, hope it's the right way. Generator on its way, working on it.


docsets/Carbon/docset.json
```diff
@@ -1,14 +1,28 @@
...
         "PHP Carbon DateTime"
+    ],
+    "specific_versions": [
+      {
+        "version": "1.31.0",
+        "archive": "versions/Carbon-1.31.0.tgz"
+      },
+      {
+        "version": "1.30.0",
+        "archive": "versions/Carbon-1.30.0.tgz"
```
Kapeli: You need to actually add this archive. Moving the `txt` file doesn't do anything.

docsets/Carbon/docset.json
```diff
@@ -1,14 +1,28 @@
...
+      },
+      {
+        "version": "1.21.0",
+        "archive": "versions/Carbon-1.21.0.tgz"
```
Kapeli: You need to actually add this archive. Moving the `txt` file doesn't do anything.

docsets/Carbon/docset.json
```diff
@@ -1,14 +1,28 @@
 {
     "name": "Carbon",
-    "version": "1.30.0",
-    "archive": "Carbon-1.30.0.tgz",
+    "version": "1.31.0",
+    "archive": "versions/Carbon-1.31.0.tgz",
```
Kapeli: This can't point to an archive within `specific_versions`. It needs to be an archive of its own, even if it's a duplicate.

docsets/Carbon/docset.json
Max13: These .tgz already exists, I had already submit them and they was processed. Does it mean, if I set specific_versions key, I have to resubmit the tgz? (Aren't they still in the CDN?)

Kapeli: They are, but they're separate. Think of the main version and the specific_versions as completely separate. You don't need to re-add the archives once they were processed at that specific_version path, but if the path changes you need to re-add them.

Kapeli:
Thank you! 👍

docsets/Carbon/docset.json
Max13: It should be fine now. My last commit sort it all out

PR #1950, opened 2018-06-26
Faker 1.7.1

Max13:
The docset was generated manually (Explained in README)


docsets/Faker/docset.json
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "Faker",
+    "version": "1.7.1",
+    "archive": "versions/Faker-1.7.1.tgz",
```
Kapeli: This can't point to an archive within `specific_versions`. It needs to be an archive of its own, even if it's a duplicate.

Max13:
@Kapeli I've amended my commit. It should be fine

Kapeli:
Thank you! 👍

PR #1951, opened 2018-06-27
Framework7 2.3.1

Max13:


docsets/Framework7/docset.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "Framework7",
+    "package": "f7",
+    "version": "2.3.1",
+    "archive": "versions/Framework7-2.3.1.tgz",
```
Kapeli: This can't point to an archive inside `specific_versions`. Add it separately somewhere, even if it's a duplicate.

Kapeli:
Failed validation:

```
Checking Framework7/versions/Framework7-2.3.1.tgz...
Empty name found for row id "329" at "Framework7.docset"
```

This means that your docSet.dsidx contains an entry without a name.

docsets/Framework7/docset.json
Max13: Sorry again for that, it's now fixed

Max13:
I've removed the empty line by hand from the docSet.dsidx. Apparently there is an empty <h*> in the doc...

Kapeli:
Thank you! 👍

PR #1957, opened 2018-06-30
Upgrade graphql to June-2018

runar-rkmedia:


docsets/graphql/docset.json
```diff
@@ -1,6 +1,6 @@
 {
-    "name": "GraphQL Specification",
-    "version": "1.0.0",
+    "name": "GraphQL June 2018 Edition",
```
Kapeli: Please don't include the date in the name. You could add this as the version. Have the name set to `GraphQL` and the version set to `June 2018 Edition` or something like that.

Kapeli:
`icon.png` needs to be 16x16.

PR #1965, opened 2018-07-09
Updating boto3 Docset to v1.7.52

angrychimp:
Additionally updating the README to provide better instructions
for managing updates.


Kapeli:
Did you forget to update the `docset.json` or is it intentional to leave the version there the same?

angrychimp:
@Kapeli - no, that was an oversight. I clearly missed that file when staging my commit. Thanks for pointing that out.

Kapeli:
Thank you! 👍

PR #1967, opened 2018-07-10
Update CUDA docset to version 9.2.148

htower:


docsets/CUDA/docset.json
```diff
@@ -1,12 +1,12 @@
 {
     "name": "CUDA",
-    "version": "9.0",
+    "version": "9.2.88",
     "archive": "CUDA.tgz",
```
Kapeli: You need to update the archive at this path as well.

htower: Ok, fixed.

Kapeli:
Thank you! 👍

PR #1986, opened 2018-07-24
add tensorlayer docset

wuhuanyu:
Add Tensorlayer docset


Kapeli:
Please don't delete the sample docset.

Also please rename everything from `Tensorlayer` to `TensorLayer`.

PR #2004, opened 2018-08-06
Fix envoy docset force-update failure

ruoshan:
@Kapeli Did I have to adjust the "specific_versions" fields accordingly to force updating the docset? I can't download the "new" version.


Kapeli:
No, `specific_versions` docsets never get updated, so doing that doesn't do anything.

The issue is that I didn't get a chance to process the docsets, I just accepted your pull request. I've done it now. Sorry for the inconvenience!

PR #2011, opened 2018-08-12
Added docset for Sublime Text

malexer:
Docset for [Sublime Text 3 official documentation](https://www.sublimetext.com/docs/3/).


Kapeli:
Thank you! 👍

PR #2031, opened 2018-08-27
9.01 release with guides to widgets.

jameshale:
Inclusion of widgets guide files and updated to 9.01 release.


Kapeli:
Thank you! 👍

PR #2050, opened 2018-09-08
Elm update 2018-09-08

pdamoc:
Contains versions of the docset for both 0.18 and 0.19


Kapeli:
Thank you and sorry for the wait.

PR #2054, opened 2018-09-09
updated SDL docset to 2.0.8

superzazu:
I updated the SDL2 docset, as it has not been updated for a while. Let me know if there is any problem.


Kapeli:
Thank you! 👍

PR #2057, opened 2018-09-15
Update Pandoc docset to version 2.2.3.2

vbwx:
Since the API, markup syntax & command line options have changed significantly in 2.x, it is necessary to keep the latest 1.x docset version.


Kapeli:
Thank you! 👍 

One small issue: in `specfic_versions`, the latest versions should be ordered towards the top. I'll fix it.

PR #2062, opened 2018-09-19
Update Vaadin to version 8.5.2

ewall:
Minor version update.


ewall:
Um, nope... something wrong with the compress bundle, it's basically empty...

PR #2071, opened 2018-09-25
Update PeopleTools to 8.57

jkrehm:

PR #2072, opened 2018-09-25
Update PeopleTools to 8.57

jkrehm:
Ref #2071

Re-submitting with compression done using Windows Linux Subsystem (instead of in Windows). I tried decompressing on my server that runs Fedora and it decompressed correctly, so hopefully Travis likes it this time.


Kapeli:
Thank you! 👍

PR #2091, opened 2018-10-10
update rxjs 6.3.3

steve3d:
RxJS has released 6.x version for a very long time. so this is the 6.3.3 version of doc.


steve3d:
and, should I create a new docset instead of overwrite the existing rxjs 5.x?

because my method only builds docset for rxjs 6.x

Kapeli:
> and, should I create a new docset instead of overwrite the existing rxjs 5.x?

I prefer if you overwrite. You can add RxJS 5.x as a specific version if you want.

PR #1817, opened 2018-03-26
Update RFCs docset to v1.5.0

weppos:
I updated the RFC docset, and it passed the validation using the tool mentioned at #1051.

You can download the archive at
https://cl.ly/0s1k080e1N0K/download/RFCs.docset-v1.5.0.tgz

/cc @willnorris

PR #2094, opened 2018-10-12
Update RFCs docset

FiloSottile:
You can download the archive at
https://dl.bintray.com/filosottile/generic/RFCs.docset-v1.6.0.tgz

/cc @willnorris, @weppos


Kapeli:
`icon.png` needs to be 16x16 in size. You also need to add `icon@2x.png` which needs to be 32x32 in size.

weppos:
Ciao @FiloSottile, thanks! Let's get https://github.com/willnorris/rfcdash/pull/9 sorted out and then I will help with this PR.

weppos:
@Kapeli out of curiosity, can we skip the icon as it has been so far? There is really no special applicable icon here.

See https://github.com/Kapeli/Dash-User-Contributions/pull/1817. In the past I intentionally left it with no icon, and the default one was used.

Would you be ok with that?

Kapeli:
@weppos sure, icons are not required, but if they're added they need to meet the requirements.

weppos:
@FiloSottile I don't think I can edit your PR. Do you mind to remove the change to the icon?

Kapeli:
I can merge and remove the icon. Is that ok @FiloSottile?

FiloSottile:
Deleted the icon, sorry for the radio silence!

Kapeli:
Thank you! 👍

PR #2108, opened 2018-10-25
Description of changes

jameshale:
Based on LC 9.01 Stable - syntax coloring and links for associations.

### DIFF
--- a/docsets/LiveCode/docset.json
+++ b/docsets/LiveCode/docset.json
@@ -1,6 +1,6 @@
 {
     "name": "LiveCode",
-    "version": "1.6.1",
+    "version": "1.7",
     "archive": "LiveCode.tgz",
     "author": {
         "name": "James Hale",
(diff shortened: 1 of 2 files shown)


docsets/LiveCode/docset.json
```diff
@@ -10,12 +10,12 @@
...
-             "version": "1.6.1",
-             "archive": "versions/1.6.1/LiveCode.tgz"
+             "version": "1.7",
+             "archive": "versions/1.7/LiveCode.tgz"
          },
         {
-             "version": "1.6",
-             "archive": "versions/1.6/LiveCode.tgz"
+             "version": "1.6.1",
+             "archive": "versions/1.6.1/LiveCode.tgz"
```
Kapeli: You need to add the docset at this path.

jameshale: ?
There wasn't anything new here.
I wanted to delete the 1.6 docset and leave the 1.6.1 docset.
Rather than add the new text in the json for the 1.7 docset I just edited the existing text...
The only new addition was the 1.7 docset.

Previously I just added new docsets and new json entries but have never added more than one docset at a time.

I thought this json edit would leave the 1.6.1 docset as it was, delete the 1.6 and add the new 1.7.

What have I got wrong?

Kapeli: Argh. Sorry. You're right. I was confused by the diff and for some reason I thought the 1.6.1 version was new, but it's already there.

Kapeli:
Thank you! 👍

PR #2116, opened 2018-11-02
Update aiohttp docset to 3.4.4

laggardkernel:
- Set icons' background as transparent
- New detailed guide for docset generation
- Bundle the aiohttp tutorial named aiohttp-demos into the docset


Kapeli:
Thank you! 👍

PR #2126, opened 2018-11-09
Add TCLAP 1.2.2

johnbeard:
Hopefully I did this right - it's the first one I've tried!


Kapeli:
This isn't a valid docset. As far as I can tell you're using the Dash index format, but you're missing the `isDashDocset` key in Info.plist.

Issue #2132, opened 2018-11-13
Vulkan docs extremely slow


warmwaffles:
The current vulkan docs are extremely slow due to it being one big html file.

Instead it should probably use the https://www.khronos.org/registry/vulkan/ the  [API Reference Pages](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/)

Kapeli:
//cc @antiagainst

antiagainst:
Yeah, originally the Vulkan spec does not have a really good chunked version. I think that gets better now. I can look into switching to the chunked version.

warmwaffles:
@antiagainst that would be awesome.

PR #2137, opened 2018-11-19
Changed to use chuncked Vulkan spec and update to version 1.1.93

antiagainst:
Fixes https://github.com/Kapeli/Dash-User-Contributions/issues/2132


Kapeli:
Thank you! 👍

PR #2180, opened 2019-01-08
Update Discord4J docset to 2.10.1

jammehcow:
Minor change, still updated.
Checked and works fine.


docsets/Discord4J/docset.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "Discord4J",
-    "version": "2.10.0",
+    "version": "2.10.1",
     "archive": "Discord4J.tgz",
```
Kapeli: You need to update this archive as well, not just the specific_version one.

jammehcow: Added!

Kapeli:
Thank you! 👍

PR #2187, opened 2019-01-17
Enable JavaScript on SuperCollider docSet

edwardloveall:
I missed that in this version, the examples in the documentation are loaded by a local JavaScript file. This is the same version of the docset, just with JavaScript enabled.

Let me know if you'd also like me to update the plist that points to the version. I wasn't sure what the protocol was here. Thanks!


Kapeli:
Thank you!

PR #2200, opened 2019-01-31
Added Agora Objective-C Docset

jiewang2:


Kapeli:
Looks like this docset is not indexed?

Also the `icon.png` is not `16x16`.

wangjie-agora:
> Looks like this docset is not indexed?
> 
> Also the `icon.png` is not `16x16`.

How should I index the docset? I've specified the index in the Info.plist by adding this key:
	<key>dashIndexFilePath</key>
	<string>docs/headers/Agora Objective-C API Overview.html</string>
Is this not working?

Kapeli:
Please see 12.4 and 12.5 at https://kapeli.com/docsets#createsqlite.

PR #2205, opened 2019-02-02
Perl6 docset

schmidh:
Perl 6 docset has not been updated for years. It is obsolete. Old scripts never worked because of wrong folders and typos in scripts. Old scripts were obviously a one-shot. Documentation has never been updated.


schmidh:
CI is stating that checks have failed. The problem is all these files are there. CI has problems with file names that are in different capitalized versions. E.g. file "type/Int.html" is at its place but there is also a "int.html" file. This somehow trips up CI.

Kapeli:
You need to fix that. Docsets are used both on case-sensitive and case-insensitive file systems. You can't have both "Int.html" and "int.html".

PR #2208, opened 2019-02-04
Update Zsh to 5.7.1

marcusadair:
It's my first time doing this, using the same script written by the maintainer (pinging @zmwangx), and so I left the README as is.

Is it ok that I put the old Zsh.tgz.txt into Zsh/versions/5.4.2/, or does it need to be the complete old Zsh.tgz file?

Thanks


Kapeli:
You need the complete old Zsh.tgz. Moving the txt file around doesn't do anything.

Kapeli:
Thank you! 👍

PR #2209, opened 2019-02-04
Updating boto3 docset to version 1.9.86

angrychimp:
Plus a minor tweak to the README since I keep forgetting about that step.


Kapeli:
Thank you! 👍

PR #2228, opened 2019-02-08
updates Perl6 docset for February 2019.

schmidh:


schmidh:
There will be monthly updates for Perl6 from now on. Hope this is ok for you.

See the discussion on [Perl 6 Doc](https://github.com/perl6/doc/issues/1185#).

Kapeli:
Thank you! 👍 

Frequent updates are fine. Even once a week is fine, if you think it's necessary. More often than a week might be annoying to users due to the update process in Dash (which involves showing a notification to the users).

Please note that I don't manually check all docset updates, so don't rely on me catching bugs. I only manually check new docsets or when it feels like something major has changed, or the contributor is not the regular maintainer of the docset. I usually just check that the CI passed and merge.

PR #1914, opened 2018-06-02
Add Jest@23.1.0

chinesedfan:
Related to #1569.

I don't know how to add 2 authors, so forgive me to replace the original author @epitaphmike. The generating scripts are based on https://github.com/epitaphmike/jest-dash with some updates.

Main differences of the docset are,
- Fetch documents from the current Jest website
- Entry types are more than `Section`
- Support to search API and configuration options now, https://github.com/epitaphmike/jest-dash/issues/2
- Keep page header/footer and side menu the same with the website

PR #2237, opened 2019-02-10
Add Jest@24.1.0

chinesedfan:
Changes:

- Exclude other languages, old versions and blogs(makes the tgz reduce to 1.28MB, compared with 8.95MB for 23.1.0 in #1914)
- Revert to remove header again, and fix the padding top
- Fix absolute links to relative links, so that it will not navigate to the real website any more
- Update README.md with the new repository


Kapeli:
Thank you! 👍

PR #2251, opened 2019-02-15
Update PeopleTools 8.57 to revision 2

jkrehm:
I am on Windows and use Velocity there. I tested the docset and I believe it's working correctly, but I cannot test the installation part. In the past the archive had problems on *nix machines, so I built it using WSL (Windows Sub-system for Linux). Hopefully it works correctly. 🤞


Kapeli:
Thank you! 👍

PR #2259, opened 2019-02-21
add carthage cheat sheet docset

mltbnz:
Adds a cheatsheet for carthage dependency manager https://github.com/Carthage/Carthage


Kapeli:
Cheat sheets should be contributed at https://github.com/Kapeli/cheatsheets.

PR #2261, opened 2019-02-25
Add SWI Prolog docset v8.1

mye:


Kapeli:
The docset archive contains "SWI-Prolog.app"? It should only contain the docset.

The icon must be `16x16` in size and you also need to add an `icon@2x.png` which must be `32x32` in size.

Please follow the contribution guidelines.

PR #2262, opened 2019-02-26
Octave v5.1.0

obstschale:
- [x] Main Docset replaced
- [x]  `docset.json` updated with new version
- [x]  Add version Docset
- [x]  Update Readme


Kapeli:
The docsets seem to not be indexed. They're missing their `docSet.dsidx`.

PR #2279, opened 2019-03-05
Update Discord4J docset to 3.0.0

jammehcow:


Kapeli:
The docsets don't seem to be indexed.

PR #2306, opened 2019-03-27
Adde Agora Web 2.6

wangjie-agora:
Update the docset to 2.6


Kapeli:
See https://github.com/Kapeli/Dash-User-Contributions/blob/master/Sample_Docset/docset.json#L14-L26 for an example on how to add specific versions. Adding just the files is not enough.

Kapeli:
Thank you! 👍

PR #2321, opened 2019-04-08
Chainer V5.4.0 docset

belltailjp:
This is to update the Chainer (a deep learning framework) docset to the latest v5.4.0.

CC: @mitmul


Kapeli:
Thank you! 👍

PR #2338, opened 2019-04-27
Add a DocSet for CQL v3.1

theckman:
There is already a docset for CQL v3.3, under the `CQL` name, with this creating
a new DocSet for v3.1 (Cassandra 2.0 and 2.1).

Signed-off-by: Tim Heckman <t@heckman.io>


kapeli:
Please add this as a [`specific_version`](https://github.com/kapeli/Dash-User-Contributions/blob/master/Sample_Docset/docset.json#L14-L25) of the existing CQL docset.

theckman:
@Kapeli so the reason I wanted to go this route was so that I could have an alias in Alfred to search for CQL 3.1 docs and not 3.3. Is there another way to do this without a dedicated docset?

Kapeli:
The `specific_versions` just tells Dash where to show the docset in Preferences > Downloads. After installation, the alias and so on come from the `Info.plist`, which you can set to be different from the 3.3 docset.

theckman:
@Kapeli my goal was to make it so I, and my coworkers, could get both 3.1 and 3.3. My goal of this addition was to get both on my system without manual toil. Is there a way I can make that automatically happen for me across my multiple machines, as well as for my colleagues?

Kapeli:
With specific_versions you can install both.

Use
https://kapeli.com/dash_install or sync files to automate.

theckman:
@Kapeli it's still not clear to me, without making them separate docsets, how someone can have an Alfred trigger like `cql` that works for both 3.1 and 3.3 docsets.

Are you able to elaborate further on that?

Kapeli:
See attached screenshot on how `specific_versions` show up in Dash. Each of the "Download" buttons installs a separate docset, which you provide within `specific_versions`. That docset can have a different docset keyword, using the `DashDocSetKeyword` key in the docset's Info.plist.

Screenshot 2020-03-02 at 10 36 13

PR #2353, opened 2019-05-06
Add Laravel Zero documentation

godbout:
The docset has icons, index page, TOC, CSS is tweaked for Dash, menu, sidebars and all removed. I think it's pretty solid.


godbout:
Ok, seems my 2x icon is not OK. Will update.

godbout:
Ok, seems not having a 2x icon is not OK either. Add a new beautiful one.

Kapeli:
Thank you! 👍

PR #2361, opened 2019-05-14
Update Laravel Zero

godbout:
Sorry, there is a mistake in the documentation itself. I've contacted the author but didn't have any answer yet. Still, because of that mistake, the version number I got from the executable is wrong. This is the updated version.

I'm not sure if this would have any consequence, so I didn't remove the 2.2.0 version. But if it's fine, we can delete it.

Thank you!


godbout:
Also, added much better Guides and Sections.

godbout:
There's an issue with my archiving script, sorry about that. Will update with a valid version. Still learning the process :))

godbout:
I did not touch the 2.2.0 archive, not sure why it shows as changed in the commit. Can be removed probably anyway, as asked above.

Kapeli:
Thank you! 👍 

Regarding 2.2.0 archive: once I merge something I rewrite the git history and remove all archives (to keep the repo size small), so the 2.2.0 archive was removed and probably came back from your repo, that's why it's showing as changed.

godbout:
@Kapeli oops, my bad. Would it be ok if I remove that `2.2.0` version completely, or that might break something on the user's side?

Kapeli:
Nope, it's ok to remove stuff.

PR #2362, opened 2019-05-14
Add Tailwind CSS v1.0 Documentation

godbout:


godbout:
JFI, the archive file name has changed. Different versions have different names. Hope it doesn't cause issue. The `docset.json` has been updated accordingly.

docsets/Tailwind_CSS/docset.json
```diff
@@ -1,13 +1,20 @@
 {
   "name": "Tailwind CSS",
-  "version": "0.7.4",
-  "archive": "Tailwind_CSS.tgz",
+  "version": "1.0",
```
bkbooth: Nitpick but I think it should be version 1.0.0 (there's also a 1.0.1 version already, probably doesn't affect the docs much though)

godbout: hey @bkbooth, yeah, I considered both cases. The reason I chose the `1.0` model is because the patch number should not affect the doc (in theory), only the major and minor number should. So no gain, but a disadvantage in Dash: you'd have a litany of versions like `1.0.1`...`1.0.24` etc... Probably it would be fine to delete old patch versions, but then it still extra work for no benefit, as far as I can see.

Every time a new Tailwind version comes out (even a patched one), I'll update the doc, but without updating the versioning number.

godbout: They have a release note page too, now: https://tailwindcss.com/docs/release-notes

Will be nice to see what they do on that also.

bkbooth: Yeah, that's a really good point. I currently have that issue with the Atom docset that I maintain, I should probably change it to a similar version model.

bkbooth: Oh nice, release notes is very handy!

godbout: Some info on updating docsets without changing the version number: https://github.com/Kapeli/Dash-User-Contributions/wiki/Docset-Versioning-Guidelines

godbout: I think both models can be valid, might be just a matter of preference. Some users might prefer seeing a stricter version number to be sure they're having the latest doc. Personally I'm happy with the `major.minor` for now ✌🏼️

Kapeli:
Thank you! 👍

PR #2367, opened 2019-05-15
Add docset for latest version Ramda v0.26.1

localshred:
I've loaded this docset locally and everything appears to be working as expected. Please let me know if there's anything I can do to get this merged. Thanks!


docsets/Ramda/docset.json
```diff
@@ -1,13 +1,17 @@
 {
   "name": "Ramda",
-  "version": "0.24.1",
+  "version": "0.26.1",
   "archive": "Ramda.tgz",
```
Kapeli: You need to update this archive as well, otherwise it will still point to the `0.24.1` docset.

localshred: Ah, sorry about that. Updated.

Kapeli:
Thank you! 👍

PR #2378, opened 2019-05-22
Update Vaadin with versions 8.8.0 and 8.8.1

ewall:
New minor versions 8.8.0 (previous) and 8.8.1 (current)

### DIFF
--- a/docsets/Vaadin/docset.json
+++ b/docsets/Vaadin/docset.json
@@ -1,6 +1,6 @@
 {
     "name": "Vaadin",
-    "version": "8.7.2",
+    "version": "8.8.1.",
     "archive": "Vaadin.tgz",
     "author": {
         "name": "Eric Wallace",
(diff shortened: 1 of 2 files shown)


docsets/Vaadin/docset.json
```diff
@@ -12,6 +12,14 @@
         {
             "_comment": ""
         },
+        {
+            "version": "8.8.1",
+            "archive": "versions/8.7.2/Vaadin.tgz",
```
Kapeli: This path seems to be wrong. It points to `8.7.2` instead of `8.8.1`.

docsets/Vaadin/docset.json
```diff
@@ -12,6 +12,14 @@
...
+        },
+        {
+            "version": "8.8.0",
+            "archive": "versions/8.7.2/Vaadin.tgz",
```
Kapeli: This path seems to be wrong. It points to `8.7.2` instead of `8.8.0`.

docsets/Vaadin/docset.json
ewall: Whoops, yup... moving too fast! Will resubmit.

PR #2380, opened 2019-05-23
Update Tailwind CSS docset

godbout:
* update tailwind to latest doc (quite a few changes since last week)
* fix an issue with TOC with names containing slashes
* add resources entries
* add samples entries


Kapeli:
Thank you! 👍

godbout:
@Kapeli just wondering, I updated to the latest version through Dash but the entries list didn't update. I restarted Dash but same thing. Only when I've removed the docset from Dash and redownloaded it then the new entries appeared. Is there something that I didn't do correctly in the update? I've followed the versioning guide. Thanks.

Kapeli:
Most likely you have the docset installed from specific versions (i.e. the popover). Those docsets don't get updated, only the main docset does. In other words, adding the `/1` at https://github.com/Kapeli/Dash-User-Contributions/blob/dad8b00d51c4c0a152a02c96f2cfa085ec93ce27/docsets/Tailwind_CSS/docset.json#L15 is a no-op.

Kapeli:
If this is not the case, let me know as it's most likely a bug.

godbout:
Ah, I see. That's correct, I've installed it through the popover. Although, I got a notification that the docset has indeed been updated. So, I do understand now the issue, but the notification makes it slightly confusing.

I'll remove the `/1` at the next update. Thanks for the help!

Kapeli:
Maybe you have installed both the main docset (that gets updated) and the versioned docset (that doesn't get updated)? You would get the update notification for the main docset in that case.

godbout:
Might be. I've never bothered with this to be honest. It's just the first time I had to look around there because the first version of the docset was done by someone else and I was trying to understand how things worked. I'll keep an eye, but I should be fine, and probably Dash is fine too. If it was a bug I'm pretty sure it would have already been raised by now!

PR #2394, opened 2019-06-13
Update Atom docset to v1.38.1

bkbooth:


bkbooth:
@Kapeli I'm going to change to only maintaining specific versions for minor releases, I'll still update when patch versions are released, but I'll just update the current minor release.

I'd like to drop some older versions so I'm assuming that I can just drop some of the pre-1.x versions from the `specific_versions` list and they'll be cleaned up?

I'd also like to change the previous versions to drop the extra patch releases, is that possible? My thinking would be something like rename the latest patch release under a minor version to just the minor version and drop the rest of the patch versions for that minor version.

For example, change this:

```js
{
  // ...,
  "specific_versions" [
    // ...,
    {
      "version": "1.32.2",
      "archive": "versions/1.32.2/Atom.tgz"
    },
    {
      "version": "1.32.1",
      "archive": "versions/1.32.1/Atom.tgz"
    },
    {
      "version": "1.32.0",
      "archive": "versions/1.32.0/Atom.tgz"
    },
    // ...
  ]
}
```

Into this:

```js
{
  // ...,
  "specific_versions" [
    // ...,
    {
      "version": "1.32",
      "archive": "versions/1.32/Atom.tgz"
    },
    // ...
  ]
}
```

Kapeli:
That's fine, as long as you provide the archive at `versions/1.32/Atom.tgz`.

bkbooth:
Ah, that's what I'd wondered. I'll have to re-provide the archives if I do this refactor, there's no way to move/rename a previous archive?

Kapeli:
No way to move/rename, sorry.

I could do it manually though on my end. Just make the changes to `docset.json` and I'll take it from there. You just want to keep the latest minor versions right? So for example `1.28.2` becomes `1.28` while all the other `1.28.x` versions get deleted.

bkbooth:
I thought that'd probably be the case.

I'd really appreciate if you can do it manually, but it's fine if you don't want to, I can generate them all again when I create a PR.

And yeah, just keeping the latest patch within each minor version. I won't have a chance to make the PR with the changes to `docset.json` now, but I'll get it done sometime in the next day or so.

Kapeli:
It's ok. Shouldn't be much work. I'll wait for your PR and then change the files to match your `docset.json`.

PR #2426, opened 2019-07-06
Updated FileMaker to 18.0.0

jpscharf:


docsets/FileMaker/docset.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "FileMaker",
-    "version": "17.0.1",
+    "version": "18.0.0",
     "archive": "FileMaker.tgz",
```
Kapeli: You need to update the main archive as well.

jpscharf: Sorry about that. I updated the main archive, rebased, and squashed to a single commit.

Kapeli:
Thank you! 👍

PR #2440, opened 2019-07-16
Update TensorFlow docs to 1.14.0

reuben:
This docset was built using a different recipe than the previous versions available. Previous versions by Yuxin Wu used this method: https://github.com/ppwwyyxx/dash-docset-tensorflow

The old recipe requires scraping the TensorFlow website, which I couldn't complete due to the server blocking my address. I developed [an alternative solution](https://github.com/reuben/gen_tf_docset) that renders the official Markdown into HTML instead of scraping the official HTML. While it doesn't match the looks of the docs in tensorflow.org, the content is complete and faithful. I've also added syntax highlighting for code snippets.


reuben:
Looks like I forgot to deal with files with the case-sensitiveness problem in the generated archive.

reuben:
Green tests, looks like the fix worked. (I appended a _{counter} to conflicting files when generating the HTML: https://github.com/reuben/gen_tf_docset/commit/f48eb0fc6dcdf42605fca058bf3ac16a40c71a0b).

Kapeli:
Thank you! 👍

PR #2451, opened 2019-07-26
Add a docset for the Click python cli toolkit

rcoup:
http://click.palletsprojects.com


rcoup:
CI says:
```
Checking Click/Click.tgz...
File not found at "py-modindex.html" for entry with row id "271" of "Click.docset"
```

@Kapeli know what this means?

(is via doc2dash, and seems to _work_ file locally...)

Kapeli:
That error means that one of the entries in your docSet.dsidx points to a file that does not exist.

Search for "module" within the docset and you'll notice the entry "Module Index" is a 404.

PR #2456, opened 2019-07-30
update shapely

ppwwyyxx:


docsets/shapely/docset.json
```diff
@@ -1,10 +1,20 @@
...
     },
     "aliases": [],
+		"specific_versions": [
+			{
+				"version": "1.6.4",
+				"archive": "versions/1.6.4/shapely.tgz"
+			},
+			{
+				"version": "1.5.15",
+				"archive": "versions/1.5.15/shapely.tgz"
```
Kapeli: You need to add the archive at this path as well.

ppwwyyxx: Thanks! Now added.

Kapeli:
Thank you! 👍

PR #2472, opened 2019-08-19
Add PyQt5 5.13.0 docset

yshalsager:


Kapeli:
Your docset's docSet.dsidx contains a lot of entries with new lines in their names. This is not currently supported. Please fix and resubmit.

PR #2477, opened 2019-08-23
Added Symfony Docs https://symfony.com/doc

pouyana:
Adds Symfony documentation guides from https://symfony.com/doc for all supported Symfony versions.

### DIFF
--- a/docsets/Symfony_Docs/README.md
+++ b/docsets/Symfony_Docs/README.md
@@ -1,0 +1,17 @@
+Symfony Docs Docset
+=======================
+
+This Docset was created by [Pouyan Azari](https://github.com/pouyana)
+
+The documentation is available as [source code](https://github.com/symfony/symfony-docs) and [online](https://symfony.com/doc).
+
+The Docset was generated with [symfony-docset-gen](https://github.com/uniwue-rz/symfony-docset-gen), please report errors und bugs there.
+
+This documentaion is valid for the following versions of Symfony:
+
+- 2.8
+- 3.4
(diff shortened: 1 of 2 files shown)


Kapeli:
The docSet.dsdix contains an entry that points to "https://github.com/bitprophet/alabaster". Is this intentional?

Also you added an archive for `versions/5.0/Symfony_Docs.tgz`, but didn't add any `specific_version` that links to it.

pouyana:
Hi indeed it shouldn't be added. The link was inside the index file, I should make the parser ignore the link and don't add it to the sqlite database. I will try to correct it.

About the version: 5.0 is the most recent version, should the most recent version be added to the versions folder or the file in root dir is enough?

I also saw there is an official Symfony API docset, is there a possibility to add these guides to that so all documentation are in one place?

Kapeli:
> About the version: 5.0 is the most recent version, should the most recent version be added to the versions folder or the file in root dir is enough?

If you add the archive to the versions folder, you also need to mention it in `specific_versions` in the docset.json. Otherwise, you only need the latest version in the root dir. Either way is fine, it's up to you.


> I also saw there is an official Symfony API docset, is there a possibility to add these guides to that so all documentation are in one place?

I haven't seen any demand for this yet (no user requested the guides). I'll watch how popular your docset becomes and I'll consider including it into the main Symfony docset based on that.

pouyana:
So I think this should work, thanks for waiting :).

Kapeli:
Thank you! 👍

PR #2491, opened 2019-09-01
Update Docset for AWS CloudFormation Template Reference 2019.08.29

majwhite:
Updated Docset to AWS CloudFormation Template Reference 2019.08.29


Kapeli:
Thank you! 👍

PR #2497, opened 2019-09-04
added google cloud client for python 0.34.0

davydhong:


Kapeli:
What's the purpose of the stuff in `docsets/Alembic`?

`icon.png.png` should be called just `icon.png` and `icon@2x.png.png` should be `icon@2x.png`.

PR #2503, opened 2019-09-14
Add Chassis 3.0.0. docset

BronsonQuick:
This adds [Chassis](http://docs.chassis.io/en/latest/) docs.

Thanks! 🙇


Kapeli:
Your docset's `docSet.dsidx` seems to be empty?

BronsonQuick:
@Kapeli Ahh right. I just looked at the package contents that doc2dash generated and it is empty. I though that it would generate the index for me. Looks like I might need to update the Sphinx templates. I don't supposed you might now of any other user generated Dash docs that go from Sphinx -> docs2dash so I can grep their codebase to see what I'm missing. I'll also log an issue upstream to see if I can get some more information on if docs2dash is supposed to generate an index for me or not.

Failing that I'll follow you [instructions](https://kapeli.com/docsets#fillsqlite) and code something up myself. Thanks very much!

Kapeli:
[A lot](https://github.com/Kapeli/Dash-User-Contributions/search?q=sphinx&unscoped_q=doc2dash) of the docsets here are made by doc2dash. Most likely you're missing a genindex.html file which doc2dash uses to generate the index.

PR #2506, opened 2019-09-15
Add Chassis 3.0.0 docs

BronsonQuick:
This adds [Chassis](http://docs.chassis.io/en/latest/) docs.

@Kapeli I reverse engineered doc2dash and decided to use `py:data::` markup in Sphinx to generate a basic index through doc2dash even though our project isn't Python. This should be enough to get this into Dash. 🤞 Thanks!


Kapeli:
Thank you! 👍

PR #2507, opened 2019-09-15
updated SDL docset to 2.0.10

superzazu:
I just updated the SDL docset to 2.0.10. Let me know if there is any problem.


docsets/SDL/docset.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "SDL",
-    "version": "2.0.8",
+    "version": "2.0.10",
     "archive": "org.libsdl.sdl20.docset.tgz",
```
Kapeli: This archive does not exist. You added `org.libsdl.sdl20.tgz` instead.

superzazu: Thank you, I fixed the typo.

Kapeli:
Thank you! 👍

PR #2513, opened 2019-09-17
Add PySide2 5.13.0 docsets

marimeireles:


Kapeli:
The [author](https://github.com/Kapeli/Dash-User-Contributions/blob/master/Sample_Docset/docset.json#L5-L8) in docset.json is required. You can use any name you want or your GitHub username, but something must be there.

marimeireles:
Hey @Kapeli, thanks, sorry for not adding it.
Would you rather me to make only one commit instead of these two?

Kapeli:
Thank you! 👍

Issue #2254, opened 2019-02-18
Request: Godot


thyrgle:
Docs:

https://docs.godotengine.org/en/3.0/

vonagam:
Closed by #2774

PR #2526, opened 2019-09-29
Godot Docset

kassane:
Reference: [this request](https://github.com/Kapeli/Dash-User-Contributions/issues/2254)


Kapeli:
Failed validation:

```
Checking Godot/Godot.tgz...
File not found at "classes/class_@c" for entry with row id "1817" of "godot.docset"
File not found at "classes/class_@c" for entry with row id "1818" of "godot.docset"
Found directory in main index: Entry with row id "5293" and path "" at "godot.docset"
Found directory in main index: Entry with row id "5294" and path "" at "godot.docset"
```

The "Found directory" is wrong. The error should be that you have entries with empty paths in your docSet.dsidx.

PR #2528, opened 2019-09-29
Poco Library v1.9.4

kassane:


docsets/Poco/docset.json
```diff
@@ -1,14 +1,18 @@
...
         "name": "wxk6b1203",
         "link": "https://github.com/wxk6b1203"
     },
     "aliases": ["poco", "pocoproject"],
     "specific_versions": [{
-         "version": "1.9.0",
-         "archive": "versions/1.9.0/Poco.tgz",
-    }]
+         "version": "1.9.4",
+         "archive": "versions/1.9.4/Poco.tgz"
```
Kapeli: You didn't add an archive at this path. I think you added it at `versions/Poco.tgz` instead.

PR #2533, opened 2019-10-01
XMake Fixed

kassane:
Img assets fixed;
Zh-cn support


Kapeli:
You can't add a combined English and Chinese docset. If you want, you can add a separate Chinese-only docset.

PR #2555, opened 2019-10-21
add a new version of Python_zh_cn

bokix:
add v3.8.0 documents.


docsets/Python_zh_cn/docset.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "Python (Chinese)",
-    "version": "3.7.4",
+    "version": "3.8.0",
     "archive": "Python_zh_cn.tgz",
```
Kapeli: You need to update this archive as well.

Kapeli:
Please resubmit when ready.

PR #2557, opened 2019-10-23
Fix Atom docset v1.40.1

bkbooth:
- Fixed broken internal doc file links from changed file/folder layout
- Enable grouping of methods in Dash search


Kapeli:
Thank you! 👍

PR #2561, opened 2019-10-28
add a new version of python document(chinese version)

bokix:
add new file in folder: /Python_zh_cn/ , it's same as file in "/Python_zh_cn/versions/3.8.0/"


Kapeli:
Thank you! 👍

PR #2562, opened 2019-10-28
Add docset for Unicode Characters 12.1.0

remko:
Added a docset containing all unicode characters from the unicode spec.


Kapeli:
Thank you! 👍

PR #2566, opened 2019-11-02
Disctionart updates from 9.5.1rc1

jameshale:
Significant additions/mods to dictionary.
Syntax colouring finally back.


Kapeli:
Thank you! 👍 

I just noticed something: you're using a separate version for the docset than the LiveCode API? Generally, I recommend that docsets have the same version as the API they're for, so when you install a docset you know for which version of the API it's for.

If this is not possible, I'd recommend starting the version with a `/`, which will have the effect of hiding it from the users.

PR #2582, opened 2019-12-01
openFrameworks 0.11.0 added

icq4ever:


docsets/openFrameworks/docset.json
```diff
@@ -9,6 +9,10 @@
     "aliases": ["of"],
 
     "specific_versions": [
+        {
+			"version": "0.11.0",
+			"archive": "versions/0.11.0/openFrameworks.tgz"
```
Kapeli: You need to add the archive at this path as well. Did you add it to `versions/0.10.1/openFrameworks.tgz` by mistake?

icq4ever: sorry I forgot that.
just put new archive in 0.11.0

Kapeli:
Thank you! 👍

PR #2615, opened 2020-01-10
Update date-fns docset to 2.8.1

jariz:
Updates date-fns to 2.8.1.

Also filed a PR on iansinnott/date-fns-dash-docset#1 but original author isn't replying, so thought I'd just upstream this myself now.


Kapeli:
Thank you! 👍

PR #2626, opened 2020-01-20
update altair docset to 4.0.0

hoishing:


docsets/altair/docset.json
```diff
@@ -1,14 +1,21 @@
...
+    "aliases": ["altair",
+                "altair 4.0"],
+    "specific_versions": [
+      {
+        "version": "4.0.0",
+        "archive": "versions/4.0.0/altair.tgz",
+      },
+      {
+        "version": "3.2.0",
+        "archive": "versions/3.2.0/altair.tgz",
```
Kapeli: You need to actually add this archive. Moving the `txt` file does nothing.

hoishing:
added tgz for the main and old specific version. pls see if the format is correct or not. thanks ~

hoishing:
oh.. seems I can't just download the old version 3.2 docset and upload its tgz ... it can't pass the CI tests.
What is the proper way if I wanna keep the old version created by others as well?

Kapeli:
You can grab the old docset from https://kapeli.com/feeds/zzz/user_contributed/build/altair/altair.tgz

Kapeli:
Thank you! 👍

PR #2634, opened 2020-01-24
Update WPILib

TheMageKing:
The WPILib in this repository is years out of date from the actual WPILib.  Additonally, it appears to only include the documentation for the Java version of the library, not the C++ version.  I added both the C++ and the updated java versions.  Can you merge it?


docsets/WPILib_Cpp/docset.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "WPILib_Cpp",
```
Kapeli: Please set the `name` to "WPILib C++"

docsets/WPILib_Cpp/docset.json
```diff
@@ -0,0 +1,11 @@
...
+    "version": "2020.2.2",
+    "archive": "WPILib_Cpp.tgz",
```
Kapeli: The archive at this path contains the WPILib Java docset, not the C++ one

docsets/WPILib_Java/docset.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "WPILib_Java",
```
Kapeli: Please set the `name` to "WPILib Java"

docsets/WPILib_Java/docset.json
```diff
@@ -0,0 +1,11 @@
...
+    "version": "2020.2.2",
+    "archive": "WPILib_Java.tgz",
```
Kapeli: If possible, please set the docset's name (the `CFBundleName` within `Info.plist`) to "WPILib Java"

docsets/WPILib_Cpp/docset.json
TheMageKing: Whoops.  Guess I messed up one of my moves...
Good catch: thanks

docsets/WPILib_Java/docset.json
TheMageKing: I'll do the same for the C++ version as well (once I actually include it lol)

TheMageKing: Should I touch the other keys as well?  Eg, DocSetFeedName or Publisher?

Kapeli: Those keys aren't used by Dash. `DocSetPlatformFamily` and `CFBundleName` are the important ones for Dash.

TheMageKing:
Doxygen appears to have failed to generate an index for the C++ files... what?
I did it according to the websites directions...
Should DISABLE_INDEX be false?

TheMageKing:
Okay, so, it seems that I need docsetutil to generate an index for my docset... but I am on a Linux box, and just don't have access to that.
What do you recommend?

Kapeli:
I'm sorry, but I have no idea what you should do in this case. I only handle the Mac side of things.

Kapeli:
I'm sorry, but no. I can't commit to doing this every time a new version is released.

PR #2636, opened 2020-01-27
add discord.js docset

n3oney:


Kapeli:
Please add a `specific_version` entry in the `docset.json` for `versions/11.5.1/discord.js.tgz`, or remove that archive as it's not needed and not doing anything right now. See https://github.com/Kapeli/Dash-User-Contributions/blob/master/Sample_Docset/docset.json#L14-L26 for an example on how to add a `specific_version`.

Kapeli:
Thank you! 👍

PR #2648, opened 2020-01-30
Update Laravel-Zero to 6.3

godbout:
Hello there! 

Update of that docset. I'm the original maintainer of it :))

Thanks!


godbout:
Not sure whether I need to add the latest version of the docset in both the root of the folder, and in specific version. Maybe not? Wouldn't it make more sense if the latest version is only at the root? Let me know. If yes I'll update the PR.

Kapeli:
You need to add them in both, yes. Thank you! 👍

PR #2657, opened 2020-02-04
Scikit-image updated to 0.16.1

xxdavid:
I also changed instructions in README as the original one didn't work
for me.


Kapeli:
Thank you! 👍

PR #2659, opened 2020-02-05
WPILib Update: ONLY JAVA

TheMageKing:
This is just the java half of the documentation: it is still very useful on its own (and, in fact, it is what was previously here).  This should be just a straight update, although I did rewrite the readme a bit.


Kapeli:
Thank you! 👍

PR #2662, opened 2020-02-07
Fix tailwind docset + update to 1.2.0

godbout:
hey, sorry. not sure what happened in my last update last week, but it seems the docset is not correctly updated. the TOC was wrong (was supposed to have instructions now, samples, etc...)

i was going to submit an update but just saw they also released a new version. so i package everything here. hope it's gonna work this time!


Kapeli:
Thank you! 👍

PR #2668, opened 2020-02-10
Update laravel-zero docset to 6.4

godbout:
6.4 has been released right after i've updated the docset to 6.3. oops! new update then. thank you!


godbout:
btw, i'm not sure whether it's a Dash 5 feature or it was the same before, but i see the section names are now followed by a kind of category from where the section comes from. that is very cool and allows me to avoid using some categories in the section names like i used to do in my previous docset. but i can't find a way to get those automatic categories to appear in title case. any help on how to format those correctly?

see screenshot below.

Screenshot_10_02_2020__18_45

godbout:
for example, for laravel that is an official docset, the categories show up in nice title case.

Screenshot_10_02_2020__18_49

Kapeli:
Thank you! 👍

Kapeli:
> btw, i'm not sure whether it's a Dash 5 feature or it was the same before, but i see the section names are now followed by a kind of category from where the section comes from. that is very cool and allows me to avoid using some categories in the section names like i used to do in my previous docset. but i can't find a way to get those automatic categories to appear in title case. any help on how to format those correctly?
> 

It's a Dash 5 feature.

Anywhere in the "path" field, add this: `<dash_entry_titleDescription=$description>`, where `$description` is the percent-encoded value you want.

Also add this to your Info.plist to prevent the default title description behaviour you see now:
```
<key>DashDocSetDeclaredInStyle</key>
<string>originalName</string>
```

godbout:
thanks for the info!

PR #2678, opened 2020-02-23
Added Unreal Engine 4 docset

### LINKED ISSUE #2020: Request: Unreal Docset
Is it possible to add this complete documentation?
[Unreal Doc](https://docs.unrealengine.com/en-us/Programming)
[Unreal API](https://api.unrealengine.com/INT/API/)

### LINKED ISSUE #2672: Unreal Engine 4 docset
So im pretty much done creating a ue4 docset.

But I won't be able to contribute it, because I will definitely hit githubs size limit.

The size of the docset is around 4.5GB the size of the compressed tar is around 2GB.
I will most likely have to provide it to you another way (like mentioned in the readme).

Summary of docset contents:

c++ API: 1.6gb
Blueprint API (Unreal's visual scripting language): 140mb
Guides, Changelogs...: 600mb
Images: 2.1gb

There is also a PythonAPI which I didn't include because the search dropdown is already struggling with 150k CPP methods. Adding the Python API which is pretty similar to the c++ API would make the search basically useless. And the python API is experimental and only for Editor scripting and therefore not used by many people anyway.

Screenshot 2020-02-19 at 12 37 11

Even c++ and Blueprint API together already amount to a lot of pages/entries (see image).

It might make sense to distribute the Python API in an extra docset, separate from the main docs.
This might also make sense for the c++ API because not everyone developing unreal uses c++.

---

One more thing functions/logic blocks in the Blueprint Visual scripting language are referred to as "nodes" (example: https://docs.unrealengine.com/en-US/BlueprintAPI/Ability/AddGameplayCueToOwner/index.html). I currently list them under "function" in the index (see image).

But I think it would be better if a dedicated "node" type/category existed so it is a bit clearer were to find the blueprint nodes.

Since unity will also add a visual scripting language this new "node" type/category could also be used there in the future.

lolleko:
Docset archive can be obtained from here: https://drive.google.com/file/d/17efBvaWeDzdwXY8T-9kKgvpqOQxY1kDh/view?usp=sharing

Closes #2672  Closes #2020


Kapeli:
Thank you! 👍

PR #2689, opened 2020-03-13
update Pillow to 7.0.0

PowerSnail:


Kapeli:
The icon.png in PyMuPDF needs to be `16x16` in size. You also need to add an icon@2x.png which should be `32x32` in size.

PowerSnail:
Done

Kapeli:
Thank you! 👍

PR #2690, opened 2020-03-15
Kotlin 1.3.70 docset

pthariensflame:
Since @kevincianfarini hasn't kept it up.

### DIFF
--- a/docsets/Kotlin/README.md
+++ b/docsets/Kotlin/README.md
@@ -1,11 +1,22 @@
-Kotlin Docset
-=======================
+# Kotlin Docset
 
-> Who are you
-Kevin Cianfarini, kevincianfarini on Github/GitLab
+## Who are you
 
-> Complete Instructions on how to generate the docset
+**Original submitter:** Kevin Cianfarini, `kevincianfarini` on
+Github/GitLab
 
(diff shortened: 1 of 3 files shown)


Kapeli:
Thank you! 👍

princed:
@pthariensflame Thanks for keeping it up-to-date! Any plans for Kotlin 1.4? 😅 

---

Also, I'm seeing some HTML rendering issue with the docset after searching:
image

Not sure it it's a problem of Dash or the docset. cc @Kapeli

pthariensflame:
@princed I was actually planning on generating the 1.4 docset this weekend!

As for the rendering issue, I’m not sure; it doesn’t happen for me. :/

kevincianfarini:
Sorry for not keeping this up to date. Zeal hasn't received an update in a while and has fallen out of my workflow. 

However, I think you all should know that Dokka is getting a major overhaul for Kotlin documentation. I'm not entirely sure if the Kotlin docs use Dokka, but if they do the scraper will need to be updated when that's put in place. 

A more sustainable solution would be to open an issue on the Kotlin Youtrack to see if there's a better documentation caching avenue than fragile webscraping. 

Good luck!

PR #2707, opened 2020-03-29
XMake v2.3.2

kassane:


Kapeli:
Thank you! 👍 

One note: newer versions need to be sorted at the top (in `specific_versions`). I'll update it.

PR #2713, opened 2020-04-02
Edited info.plist to support online redirection

venkatasg:
Only a minor update to the transformers docset - the previous version did not support online redirection to the corresponding URL for the doc page in Dash.


Kapeli:
Thank you! 👍

Issue #2699, opened 2020-03-24
Request: QEMU Documentation


kassane:
https://www.qemu.org/docs/master/

PR #2721, opened 2020-04-08
QEMU Docset

kassane:
#2699


Kapeli:
Thank you! 👍

Kapeli:
I've decided against including this docset, as it's not properly indexed.

PR #2722, opened 2020-04-08
Clang 11 Docset

kassane:
#2715


Kapeli:
This docset isn't indexed thoroughly. It has only 61 "Function" entries, which aren't really functions. The previous version of the docset was indexed a lot better.

PR #2725, opened 2020-04-08
Dr. Memory Docset

kassane:
New docset:

Dr. Memory is a memory monitoring tool capable of identifying memory-related programming errors such as accesses of uninitialized memory, accesses to unaddressable memory (including outside of allocated heap units and heap underflow and overflow), accesses to freed memory, double frees, memory leaks, and (on Windows) handle leaks, GDI API usage errors, and accesses to un-reserved thread local storage slots.


Kapeli:
Please index the docset properly.

Please change the name in docset.json to `Dr. Memory`.

Issue #2714, opened 2020-04-02
ClojureScript docset misses namespace information


mharju:
Thanks for the contributed CLJS docs, they have been in quite an extensive use for me for a long time. However, after updating to Dash 5, the information about the namespace of the matches has disappeared. This makes it hard to know which one is the correct selection.

![image](https://user-images.githubusercontent.com/130086/78246575-0a2f4880-74f2-11ea-9b2e-569d3f8fd263.png)

mharju:
@shaunlebron if there is something I can help with to get this working, I'll be happy  to help!

shaunlebron:
1. how did the UI display it before?
2. does the official clojure docset do it better?

mharju:
> 1. how did the UI display it before?

It displayed the namespace in conjunction with the match.

> 2. does the official clojure docset do it better?

Yes, it works like expected. See the attached screenshot.

![image](https://user-images.githubusercontent.com/130086/78323523-c2a0cf00-757a-11ea-86ca-39659bf41665.png)


Thanks!

shaunlebron:
@Kapeli how might I add “hierarchy” information to the symbol index?  Like in your official clojure docs:

<img width="317" alt="Screen Shot 2020-04-03 at 7 49 05 PM" src="https://user-images.githubusercontent.com/116838/78414956-6377a480-75e4-11ea-8701-d6e5e6c31f57.png">

The SQLite index only takes `name`, `type`, and `path`.  Is the path somehow used to construct the hierarchy?

Kapeli:
That menu is built from the [table of contents](https://kapeli.com/docsets#tableofcontents). It's only for the current page, not the entire docset. It's also no longer available in Dash 5.

Dash docsets do not have any hierarchy.

shaunlebron:
@Kapeli how would you suggest attaching a namespace to the search results then?  like the following:

![](https://user-images.githubusercontent.com/130086/78323523-c2a0cf00-757a-11ea-86ca-39659bf41665.png)

Kapeli:
Within the `path` field in the SQLite index, add this anywhere: `<dash_entry_titleDescription=%@>`, where `%@` is the namespace you want, percent-encoded.

shaunlebron:
I appended `<dash_entry_titleDescription=cljs.core>` to the end of the path field as suggested, and it seems to have corrupted the paths:
[...]

PR #2732, opened 2020-04-14
ClojureScript 1.10.597 (with filename fixes for prettier search results)

shaunlebron:
Dash 2 compatible fix for: https://github.com/Kapeli/Dash-User-Contributions/issues/2714

- Prefix filename with namespace (so search results have more context)
- Remove md5 suffixes (use a single digit for case-sensitive filenames)


Kapeli:
Thank you! 👍

PR #2747, opened 2020-04-22
Altair 4.1.0 and Bokeh 2.0.1

capac:
Latest versions of Altair and Bokeh docs


docsets/Bokeh/docset.json
```diff
@@ -1,13 +1,20 @@
...
+    },
+    "aliases": ["Python interactive visualization library",
+                "Bokeh Scientific plots",
+                "web browser visualization",
+                "bokeh"],
+
+    "specific_versions": [
+        {
+            "version": "0.12.3",
+            "archive": "versions/0.12.3/Bokeh.tgz",
```
Kapeli, line 17: You need to actually add this archive. Moving the txt file doesn't have any effect.

PR #2756, opened 2020-04-27
Update Qt for Python (PySide2) to 5.14.2.1

acbetter:
fixes and changes for 5.14 doc style and update doc version
more info see [https://github.com/acbetter/qt-for-python-to-dash-docset/pull/1](https://github.com/acbetter/qt-for-python-to-dash-docset/pull/1)


Kapeli:
Thank you! 👍

PR #2757, opened 2020-04-27
Add MiniZinc docset

Agelvie:
Hi!

I wanted to provide the Minizinc modeling language documentation.

The documentation is also in Spanish and Chinese, in case that they are requested they can be added quickly.


Kapeli:
Thank you! 👍

PR #2760, opened 2020-04-28
Update maintainer for fish docset and docset itself to 3.1.1

schrodincat:
👋

I'm taking over maintaining the fish docset if that's OK. We discussed this with previous maintainer and he doesn't mind.

fish shell guys changed how they generate docs recently, so it's a complete rewrite of scripts based on Dashing.


Kapeli:
Thank you! 👍

PR #2762, opened 2020-04-29
Update fish docset to 3.1.2 and strip out footers from HTML

schrodincat:
Strip out unnecessary elements from HTML, reintroduce commands page


Kapeli:
Thank you! 👍

PR #2763, opened 2020-04-29
ClojureScript 1.10.741 (fix paths for Zeal)

shaunlebron:
Remove leading slash from path keys in the database index (e.g. `/foo/bar/` -> `foo/bar`).

Dash allows them, but Zeal does not.  Fixes https://github.com/zealdocs/zeal/issues/1223


Kapeli:
Thank you! 👍

PR #2765, opened 2020-04-30
update tailwindcss to 1.4

godbout:
hello!

new version of TailwindCSS out, so here's the new (even better done) docset.

thanks. and thanks again for Dash! awesome tool.


Kapeli:
Thank you! 👍

PR #2767, opened 2020-05-01
add Laravel Zero 7.x

godbout:
hello!

please check whether what i'm doing is possible. i want to tag major versions as specific versions through the 7.x, 8.x, etc... version scheme. but i want to be able to push new minor/patch versions when needed. hope this works. let me know if i need to do some changes to this PR. thanks!


godbout:
ok, seems there's an issue with my docset. will investigate.

godbout:
is it possible actually to call a version something like _master_ or _latest_?
i'm not really sure what's the best way to version those docsets. for two of the docsets i do, the projects themselves only keep a master of the documentation. no versioning. so i don't think i should be keep any versioning either in Dash. doesn't make much sense.

Kapeli:
Yes. This is fine. You can call a version whatever you want.

PR #2769, opened 2020-05-02
Major improvements to MiniZinc docset

pthariensflame:
Note the version number: the previous one was just entirely incorrect, so this one only _appears_ to be lower, while actually being the latest release.

I've also added a bunch of new indexed entities, cleaned up the existing ones, and improved the generation README.


Kapeli:
Thank you! 👍

PR #2774, opened 2020-05-05
Godot 3.2

vonagam:
Initially tried to do with `doc2dash` but results were unsuccessful, so made it with `Dashing` like in recent haxe pull. 

Unfortunately tgz file size is too big - 247MB, here is a [dropbox link](https://www.dropbox.com/s/djmjrfmhmgelwzd/godot.tgz?dl=0).

Strange thing is that, unlike haxe docset, dash does not show a table of contents for this one. Though links are present for that...


Kapeli:
The docset fails validation:
```
File not found at "classes/class_@c" for entry with row id "4" of "godot.docset"
```

This means an entry in docSet.dsidx points to "classes/class_@c", which doesn't exist.

The table of contents doesn't work because of an issue within Dash. I'll fix it in the next Dash update.

vonagam:
You sure you have the last version? 5 hours ago (1 hour after creating the pull) i changed dropbox link, because i saw that `ignore: ["@C#"]` was not working and should have been `ingore: ["@C#¶"]`. And when there is wrong error, indeed i see that problematic link is matched forth.

vonagam:
Ok, so reuploaded again - here is [new link](https://www.dropbox.com/s/i2wob9kbangakyx/godot2.tgz?dl=0).

Also, now will try to fix problematic page (`classes/class_@c#.html`).

Kapeli:
The new version works. Thanks!

PR #2790, opened 2020-05-14
Add Apache Beam Python SDK

sadovnychyi:
https://beam.apache.org/
https://beam.apache.org/releases/pydoc/2.20.0/

Note that they have SDK for Go and Java as well, so Python suffix is mandatory.


Kapeli:
Thank you! 👍

PR #2793, opened 2020-05-17
Update Docset for AWS CloudFormation Template Reference 2020.05.14

majwhite:
Update Docset for AWS CloudFormation Template Reference 2020.05.14


Kapeli:
Thank you! 👍

PR #2794, opened 2020-05-18
Update docset for Qt for Python 5.14.2.1

acbetter:
https://github.com/acbetter/qt-for-python-to-dash-docset/issues/2


Kapeli:
Thank you! 👍 

Please note that you need to update the `docset.json` to actually push an update to users of your docset.

PR #2798, opened 2020-05-20
update Carbon PHP for v2.34.2

claytonrcarter:
Hi, so I *think* I did this more-or-less correct, but I'm totally happy to fix up what I've missed. The docset loads in Dash, sections and searching appear to work.

Two things to point out:
- the existing user contributed version of the Carbon docs (#1948 from 2018, for v1.31.0) includes the doxygen docs for the full source API, and they are in a directory called `Carbon.docset/Contents/Resources/Documents/packagist_docset` but I didn't see how to generate or download those from the git/PR history of the Carbon docsets. As such, the existing Carbon docs include a full class/API reference while this PR only includes the HTML docs available at https://carbon.nesbot.com/docs/
- I used some simple `sed` commands to disable GA, Piwik and ads in the HTML. It's not elegant, but it does appear to work fine. I figured it was better than leaving trackers and ads in place.

cc @Max13


Max13:
@claytonrcarter Thank you for helping me and the community on this one.

Add yourself to the authors on `composer.json`

claytonrcarter:
@Max13 my pleasure! Do you have any notes or recollection on how you generated the API docs in`packagist_docset` for #1948?

Max13:
@claytonrcarter Sorry I didn't answer previously, I didn't understand you needed it. 

According to [generate.bash#L27](https://github.com/Kapeli/Dash-User-Contributions/blob/65a37a34a368c236eb5e1945f78707889f76eea5/docsets/Carbon/generate.bash#L27), I was apparently executing doxygen locally, before using dashing.

I think I forgot to include the doxygen execution and parameters in my commit. Do you need me to try and submit a PR to your fork ?

claytonrcarter:
I played around w/ running doxygen locally but I couldn't get the config right, or just wasn't doing it right. I'm unlikely to have the time to fiddle with it myself, so if you have the config and/or commands handy, it would be very helpful. Either a PR or just paste them here in a comment; whatever is easiest for you. Thanks!

Kapeli:
Thank you! 👍

Max13:
Thanks. I've sent a PR to @claytonrcarter's branch to add the API doc, he should send another PR soon 😁

PR #2799, opened 2020-05-20
Tiki (TikiWiki) docset

godbout:
hello!

i'm trying to build the docset for [Tiki](https://tiki.org/HomePage). the pages are quite heavy in JavaScript though, and some elements are rendered through JavaScript processing. it works when i view the pages in a browser, but not through Dash. for whatever reason Dash doesn't seem to load the external JavaScript files. any idea why?

you can see the issue in action in the following file included in the zipped docset: PluginTrackerList.html

i'll attach a GIF below to show the issue.

thanks!


godbout:
even better: https://share.getcloudapp.com/qGudGmnq

i added a dummy alert in an external JavaScript to check and it's confirmed. the browser shows the alert, but Dash doesn't. it just doesn't seem to load the external JS file.

Kapeli:
The docset seems to work fine for me.

Did you remove and re-add it in Dash after adding the `isJavaScriptEnabled` key in Info.plist?

godbout:
god. seems you're right 😅️

i guess it was too late yesterday. thanks! gonna fill the docset in the following days and push a final PR then. thanks!

Kapeli:
Thank you! 👍

PR #2800, opened 2020-05-21
add juce6 preview docs and try supporting multiple versions

modosc:
just wanted to make sure this was the right way to introduce versioning. i don't want the `6.0.0-preview` version to be the default, so i left the version tag at `5.4.7`. thanks!


Kapeli:
Thank you! 👍

PR #2805, opened 2020-05-21
Redo of update for Carbon PHP for v2.34.2

claytonrcarter:
After @Max13 [pointed me in the right direction](https://github.com/claytonrcarter/Dash-User-Contributions/pull/1) for generating the API docs, I was able to redo the docset generation, and I added more "types" ("Guide", "Main Page", "Classes", etc) while I was at it.

So, this PR includes the contents of #2798 but adds the API docs and additional types to mix.

Sorry for additional noise, and sorry for my learning curve. I think I'm up to speed now, though.


Kapeli:
Thank you! 👍

PR #2807, opened 2020-05-22
update tailwindcss to 1.4(.5) + make it even more offline!

godbout:
doc has been updated, and docset is cleaner now. got rid of useless JS and mostly everything is offline now (except unsplash images, at least for now).

### DIFF
--- a/docsets/Tailwind_CSS/docset.json
+++ b/docsets/Tailwind_CSS/docset.json
@@ -1,6 +1,6 @@
 {
   "name": "Tailwind CSS",
-  "version": "1.4/0",
+  "version": "1.4/5",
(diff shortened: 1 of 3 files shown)


godbout:
just realized the google fonts after commit. so two commits.

godbout:
seems there's a page missing. probably because there's two pages with the same name. will investigate.

godbout:
ok so yeah it's due to the fact that i flatten what i download. in this case there's two pages that end up with the same name. i'll have to update my docset builder to handle that, then will publish an update version of this docset.

godbout:
fixed 👍🏼️

Kapeli:
Thank you! 👍

PR #2808, opened 2020-05-23
update laravel zero to 7.2.1 + clean docset

godbout:
most importantly remove unwanted JS and CSS and docset almost completely offline! (15mb videos excluded.)


Kapeli:
Thank you! 👍

PR #2813, opened 2020-05-30
update Tiki docset

godbout:
add:
* playground
* modules
* tracker fields
* wiki syntax
* online redirection
* Dash 5 description feature


Kapeli:
Thank you! 👍

PR #2817, opened 2020-06-02
add Ploi API docset

godbout:


Kapeli:
Thank you! 👍

godbout:
hello!

the icons of the docset have been stripped away. i've checked and they are present in the archive i've submitted, but not in archive downloaded through Dash.

Kapeli:
The icons within the docset are not used when installed from Preferences > Downloads > User Contributed. You need to add the icons by following the instructions at https://github.com/Kapeli/Dash-User-Contributions#contribute-a-new-docset:

> Include a icon.png and icon@2x.png with sizes 16x16 and 32x32 or simply delete the sample icon if you don't want an icon at all

godbout:
yeah but that's my point. i have an icon.png and icon@2x.png set. like the 3 or 4 other docset i submitted, that worked fine. but for whatever reason the docset downloaded don't contain those icon and icon@2x for this docset, while they do for the other ones i submitted.

or i am missing something obvious?

godbout:
oh wait. got it 😅️ 
will do another PR then. sorry about that.

PR #2820, opened 2020-06-03
Update docset for Qt for Python 5.14.2.1

acbetter:
https://github.com/acbetter/qt-for-python-to-dash-docset/pull/4


Kapeli:
Thank you! 👍

PR #2828, opened 2020-06-06
Add Swift Package Manager docset

ThatsJustCheesy:
Index of https://docs.swift.org/package-manager/, which describes the PackageDescription API for Swift Package Manager.


docsets/Swift_Package_Manager/docset.json
```diff
@@ -0,0 +1,10 @@
+{
+    "name": "Sample",
```
Kapeli, line 2: I'll change this name to "Swift Package Manager"

Kapeli:
Thank you! 👍

PR #2838, opened 2020-06-12
add dark mode for some of TailwindCSS color palettes and tables

godbout:
currently for palettes and tables. rest will come later (some issues with inputs).

### DIFF
--- a/docsets/Tailwind_CSS/docset.json
+++ b/docsets/Tailwind_CSS/docset.json
@@ -1,6 +1,6 @@
 {
   "name": "Tailwind CSS",
-  "version": "1.4/5",
+  "version": "1.4/5-20200612",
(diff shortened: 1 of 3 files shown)


godbout:
something changed in the ci between now and May 28? i have fake folder in a path just to be able to go up one level. hasn't changed since May 28 and the build passed at the time. just out of curiosity. i can find another way to handle that.

Kapeli:
Thank you! 👍

Kapeli:
> something changed in the ci between now and May 28? i have fake folder in a path just to be able to go up one level. hasn't changed since May 28 and the build passed at the time. just out of curiosity. i can find another way to handle that.

The last change to the CI was on 04.09.2019.

godbout:
fair enough. thanks.

PR #2840, opened 2020-06-13
upgrade to 0.35.0

watsy0007:


Kapeli:
The docset is not indexed. The `docSet.dsidx` is empty.

PR #2855, opened 2020-06-29
add Bulma 0.9.x Docset

godbout:
moved the previous (old) 0.7.1 to its own specific version folder.


godbout:
the build is failing on the previous Docset. should we remove it? i downloaded it from Dash as it's the current one available, archived it, and put it in its own version folder. nothing more. version 0.7 is quite old, probably pushed before the change in ci.

Kapeli:
I'll fix 0.7.

Thank you! 👍

godbout:
alright. thanks!

PR #2857, opened 2020-06-29
juce-6.0.0 final

modosc:
juce 6.0.0 released - making this the default version now but still supporting the last 5.x release as well.

### DIFF
--- a/docsets/JUCE/docset.json
+++ b/docsets/JUCE/docset.json
@@ -1,7 +1,7 @@
 {
     "name": "JUCE",
-    "version": "5.4.7",
-    "archive": "JUCE.tgz",
+    "version": "6.0.0",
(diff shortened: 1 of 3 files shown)


docsets/JUCE/docset.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "JUCE",
-    "version": "5.4.7",
-    "archive": "JUCE.tgz",
+    "version": "6.0.0",
+    "archive": "versions/6.0.0/JUCE.tgz",
```
Kapeli, line 4: This doesn't work. The main archive needs to be at a different location from the ones in `specific_versions`. I'll fix it.

modosc: thanks! wasn't sure if it'd work or not, now i know 👍

PR #2860, opened 2020-07-01
Update of Elm 0.19 packages documentation on 2020-07-01

kraklin:
Another monthly update (after 2 month period though) of Elm packages docsets


Kapeli:
Thank you! 👍

PR #2866, opened 2020-07-04
add Alfred 4 Docset

godbout:
wondering, for the current page table of content, any chance to have/use other keyboard shortcuts? rather than the Alt+up/down. something without the up and down keys, more like Ctrl+n/p, Ctrl+j/k, etc... more common, no need to move the hand that far. thanks!


Kapeli:
Thank you! 👍 

I'll consider supporting other keyboard shortcuts. Thanks for the suggestion.

PR #2868, opened 2020-07-06
Update Vulkan docset to version 1.2.146

antiagainst:
This also fixes the issue that indices for Vulkan commnands
were missing because Vulkan spec additionally prints when
a command is available in the API spec code snippets now.


Kapeli:
Thank you! 👍

PR #2877, opened 2020-07-10
update Alfred 4 to 4.1.x, save specific versions

godbout:
sorry, Alfred released a new minor version (new features) the day after i made the Docset 😅️

so here's an update. and i will keep minor specific versions.


Kapeli:
Thank you! 👍

PR #2879, opened 2020-07-12
add ARM docset

leolovenet:


Kapeli:
Failed validation for ARM System Register: some of the entries in docSet.dsidx contain newlines, which are not allowed.

Kapeli:
Thank you! 👍

PR #2894, opened 2020-07-22
Update Presto docset to 339

knu:
Presto 338 docset was missing most function and operator entries due to the document format/structure change I missed, so I'd like to replace it if possible by adopting the second commit.  Please ignore it if it's too late.


Kapeli:
Thank you! 👍

PR #2896, opened 2020-07-23
rename Alfred 4 to Alfred, as i keep track of versions now

godbout:
coz it shows current in Dash as `Alfred 4 4.1.x`. doesn't make sense. i wanna rename this to `Alfred 4.1.x`. let's see when they release version 5 how we track this. thanks!


Kapeli:
Thank you! 👍

PR #2901, opened 2020-07-25
add Chart.js Docset

godbout:
hello!

Chart.js Docset.

btw, i've been building the Docsets with a Builder i made in PHP. it's now out as a `composer` package v1.0.0: https://github.com/godbout/dash-docset-builder

might be worth a mention somewhere in the `generate your docset` page of kapeli.com? might help other people build Docsets easily. thanks!


Kapeli:
Thank you! 👍 

dash-docset-builder is mentioned at https://kapeli.com/docsets#dashDocset. I have no idea when I added the link, but it was some time ago.

If you think it would be appropriate to mention it in some other place too, please let me know!

godbout:
ah, nice! didn't see that. i was looking at the PHP section: https://kapeli.com/docsets#phpExample
maybe it would make sense to put it there too? i see Dashing seems to be only using JSON files, but Dash Docset Builder requires some (little) PHP knowledge, as the Docset settings and tweaking is done in a PHP class. a bit more knowledge to have, but it's more flexible i think. you can download the HTML docs the way you want if needed, reformatted the way you like also, etc...

thanks!

PR #2907, opened 2020-07-29
update TailwindCSS to 1.6

godbout:
* delete all versions that TailwindCSS don't keep themselves. i'll keep copies on my own repo just in case someone needs it. will clean Dash repo and interface
* fix README link to new repo


Kapeli:
Thank you! 👍

PR #2908, opened 2020-07-30
Update Presto docsets of 338 and 339

knu:
- Update the original URLs
- Hide the intrusive fixed header
- Do not load a webfont


Kapeli:
Thank you! 👍

PR #2913, opened 2020-07-31
update PEPs

ppwwyyxx:
Not sure if the version "2020.07" is allowed.
PEPs does not have "version number" AFAIK. It's updated continuously so it seems better to use date.

### DIFF
--- a/docsets/PEPs/docset.json
+++ b/docsets/PEPs/docset.json
@@ -1,11 +1,11 @@
 {
     "name": "Python Enhancement Proposals",
-    "version": "1.0",
+    "version": "2020.07",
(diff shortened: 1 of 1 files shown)


Kapeli:
Please either add an entry in `specific_versions` ([docset.json](https://github.com/Kapeli/Dash-User-Contributions/blob/master/Sample_Docset/docset.json#L14-L26)) for `PEPs/versions/2020.07/PEPs.tgz`, or remove that archive.

ppwwyyxx:
I removed the archive. Thanks

Kapeli:
Thank you! 👍

PR #2914, opened 2020-08-01
update TensorFlow 1.15

ppwwyyxx:
The old project (https://github.com/ppwwyyxx/dash-docset-tensorflow) lacks update in the past year but now I've brought it up to date.

Compared to the alternative solution in https://github.com/Kapeli/Dash-User-Contributions/pull/2440, my approach has a more native look that follows the official website, and I also make sure the links are clickable and direct to the correct doc page.

Before:

After:

@reuben , I've updated the authors to reflect contributions in each version. Let me know if that doesn't look good to you. I can also do the same for TensorFlow2 if you like.

### DIFF
--- a/docsets/TensorFlow/README.md
+++ b/docsets/TensorFlow/README.md
@@ -1,6 +1,8 @@
 TensorFlow Docset
 =======================
 
-Built by Reuben Morais
+Built by Yuxin Wu (ppwwyyxx).
+See [dash-docset-tensorflow](https://github.com/ppwwyyxx/dash-docset-tensorflow) for instructions to generate the docset.
 
+Intermediate versions 1.14.0 and 1.15.0 are built by Reuben Morais, using a different approach.
 See [gen_tf_docset](https://github.com/reuben/gen_tf_docset) for instructions to generate the docset.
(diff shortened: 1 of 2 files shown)


reuben:
My problem with scraping their website is that it's hard to reproduce if your network fails mid-scrape (which is why I wrote the script based on the Markdown source), and it's also easy to let links to their website slip through, which makes the docs fail to work if you're offline. In this new version, for example, if you go to the `tf.placeholder` docs, then click one of the links in the "Important" section (for example the`Session.run()` link), it goes to the live website instead of the local `tf.Session.run` docs.

reuben:
It does look much better though :) Personally I don't have any issues with this approach, I just can't use it myself because scraping takes too long and fails mid-way, so if I update these docs in the future they'll revert to the simpler Markdown look.

ppwwyyxx:
Thanks for spotting the issue! I had a bug in parsing the URLs and will fix it now.

I never had problems in the past regarding scraping so I didn't know it was an issue. The downloaded htmls are actually not large, I think I would upload them to my repo as well for reproducibility

Kapeli:
Thank you! 👍

PR #2926, opened 2020-08-07
Add new docset for POSIX.1-2017

genderquery:
Note that there might be some legal issues including the docset in your
repository. I have not obtained permission from the IEEE nor The Open
Group to reproduce the specification per
https://pubs.opengroup.org/onlinepubs/9699919799/copyr.html


Kapeli:
Can you try to get permission?

genderquery:
Email sent to austin-group-permissions@opengroup.org requesting permission.

PR #2929, opened 2020-08-09
add chartjs-plugin-datalabels Docset

godbout:


godbout:
i don't seem to be able to send the user to the Online Docs through Dash with this Docset?

if i press `cmd+shift+b` it still sends me to the local one. in Dash UI the "share" icon doesn't offer the `Open Online Page` menu item. it directly opens the local one.

i didn't change anything in the plist. the only thing that differs from other Docsets is that the domain name ends in `.app` which tricks macOS in believing it's an app rather than a folder. any idea how to solve this?

Kapeli:
Thank you!

Seems to be an issue with Dash. I'll look into it and will get back to you.

godbout:
thank you!

Kapeli:
> i didn't change anything in the plist. the only thing that differs from other Docsets is that the domain name ends in .app which tricks macOS in believing it's an app rather than a folder. any idea how to solve this?

This is caused by the ".app", but it's a bug within Dash, not macOS. I'll fix it in the next version of Dash. If you'd like to fix it for all versions of Dash, you'll need to make the paths not contain `.app/`. Sorry for the inconvenience!

godbout:
no worries. thanks for the heads up!

PR #2933, opened 2020-08-13
update Chart.js, add page navigation

godbout:
* stop navigation from keyboard though, and fix links that would
* sent to online doc instead of offline doc

should be the latest update before a while.


Kapeli:
Thank you! 👍

PR #2971, opened 2020-09-10
update TailwindCSS to 1.8.3

godbout:
and add a kinda Online Redirection.

i can redirect to the online page, but not to the anchor. for whatever reason Dash doesn't seem to add the anchor on this Docset. any idea why? the other ones of my Docsets work great. thanks!


Kapeli:
Can you give me step by step instructions on how to replicate the issue? Online redirection seems to work fine for me (with anchors).

godbout:
1. remove the TailwindCSS Docset
2. quit Dash
3. open Dash
4. install TailwindCSS Docset
5. search for a section
6. press Cmd+Shift+B to open online page

you should be redirected to the correct page, but not the correct anchor. no anchor in the URL.

you can see it here: https://share.getcloudapp.com/geuzeRQv

thanks!

godbout:
Posting Syndrome. right after having posted this, i switched from Safari as my main browser to Chrome. Chrome works... Safari used to work before though.

Kapeli:
I'm able to replicate the issue using Safari. As far as I can tell, no other docset works with anchor redirection when Safari is the default browser. I'll look into it. Thanks for letting me know!

godbout:
hmm. thought it could come from my side but now not sure. i restarted, same thing. it works in Chrome, Firefox, and Brave, but not in Safari. just for this Docset. Alpine works fine, Chartjs too, etc...

Kapeli:
That's weird. The Ruby docset doesn't seem to work either with Safari... I'll look into it and will get back to you in case you need to make any changes.

godbout:
> That's weird. The Ruby docset doesn't seem to work either with Safari... I'll look into it and will get back to you in case you need to make any changes.

sure, thanks.

on my side i've tested with several of my Docsets. some with the Online Redirection through HTML, some without. they all seem to work fine.

Kapeli:
> on my side i've tested with several of my Docsets. some with the Online Redirection through HTML, some without. they all seem to work fine.

I've finally looked into it.

Safari seems to have an issue where it drops the hash between redirects.

What happens is:
1. My script (located at https://kapeli.com/dash_share) redirects to `http://tailwindcss.com/docs/overflow#scroll-vertically-if-needed`)
2. `http://tailwindcss.com/docs/overflow` redirects to `https://tailwindcss.com/docs/overflow`

During the second redirect, Safari loses the hash.

You should be able to fix it by setting the comment like this:

```html
<!-- Online page at https://tailwindcss.com/docs/overflow -->
```

I seem to have the same issue in the Ruby docset as well.

godbout:
ah good catch. i don't provide a protocol. will add `https` and see if it helps.

but the issue seems on your side right? if i paste `http://tailwindcss.com/docs/overflow#scroll-vertically-if-needed` directly in Safari, i'm redirected to the `https` version, including the anchor.

Kapeli:
I have no idea what Safari does to be honest. Maybe it drops the anchor only when there's a redirect chain?

This is the response my server gives:

```
curl -IX GET https://kapeli.com/dash_share\?docset_file\=tailwindcss\&docset_name\=Tailwind%20CSS\&fallback\=https://\&path\=tailwindcss.com/docs/overflow.html%23scroll-vertically-if-needed\&platform\=tailwindcss\&source\=tailwindcss.com/docs/overflow
HTTP/2 302 
server: nginx
date: Wed, 23 Sep 2020 08:30:27 GMT
content-type: text/html; charset=UTF-8
location: http://tailwindcss.com/docs/overflow#scroll-vertically-if-needed
access-control-allow-origin: *
access-control-allow-credentials: true
access-control-allow-methods: GET, POST, OPTIONS
access-control-allow-headers: DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type
```

Notice the "location:" redirect, it includes the anchor.

PR #2980, opened 2020-09-13
Update D to 2.093.1

RUSshy:


docsets/D/docset.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "D",
-    "version": "2.069.2",
+    "version": "2.093.1",
     "archive": "D.tgz",
```
Kapeli, line 4: You need to actually update the docset archive. Just changing the version does nothing.

Issue #2956, opened 2020-08-28
webpack4


yangxyo:
webpack4 should update

godbout:
working on latest doc, v5. coming sooooooooon.

godbout:
v4 too actually.

PR #2991, opened 2020-09-20
add webpack 4.44.x Docset

### LINKED ISSUE #818: Webpack
The webpack

godbout:
can close #818 and #2956.

this is one is done grossly. the new v5 will come later and should be better.

having the same issue as TailwindCSS, with the Online Redirection and Safari.


godbout:
i'm having an issue with the anchors' positions in that one. if i search from the search bar, then the anchors work well. if i navigate from the inner TOC, they don't. might be something on my side. sometimes i have to tweak a bit depending on the docs i'm working on, but usually i have the same result from the search bar and from the inner TOC. so not sure what's happening here.

see here: https://share.getcloudapp.com/Apuk5PLr

Kapeli:
> see here: https://share.getcloudapp.com/Apuk5PLr

This seems to be because the inner TOC links using `apple_ref` anchors, while the search index links using the anchors from the `id=` of the headings (`<h3 id="define-with-factory">`).

I'd try adding the apple_ref directly above the headings (`h3`), instead of as a child.

godbout:
i did try yeah, didn't work. when i started the first Docsets i tried several ways, the current one seemed to be the best. i tried to add it before the h3 when i saw this issue for this Docset, but didn't work. another way i've tried in the past is by adding some padding/margin, but kind of a pain. anyways, not that important. was just curious about the discrepancy. thanks!

PR #2997, opened 2020-09-24
add webpack 5 Docset

godbout:


godbout:
having a little issues with certain entries. if they contain an underscore, they don't get selected in the page inner TOC. you can see the issue here: https://share.getcloudapp.com/RBuW5WA8

am i encoding something wrongly? would be totally possible. thanks!

Kapeli:
> having a little issues with certain entries. if they contain an underscore, they don't get selected in the page inner TOC. you can see the issue here: https://share.getcloudapp.com/RBuW5WA8
> 
> am i encoding something wrongly? would be totally possible. thanks!

It's just an issue with how Dash finds the anchor to highlight. If you can, please make it so that:

```html
<a id="resourcequery-webpack-specific" name="//apple_ref/cpp/Section/__resourceQuery%20%28webpack-specific%29" class="dashAnchor"></a>
```
becomes
```html
<a id="resourcequery-webpack-specific"></a>
<a name="//apple_ref/cpp/Section/__resourceQuery%20%28webpack-specific%29" class="dashAnchor"></a>
```

I can fix it in Dash too, but it would require a Dash update and it wouldn't work with older versions of Dash.

godbout:
i can make the change, yes. if it's a bug in Dash (it does feel like it), it'd be nice to have it fixed in there too. i don't mind also people having to use the latest version of Dash. i'll update my other Docsets only when i'll get a latest doc. thanks!

Kapeli:
It's definitely a bug in Dash, yes. I'll fix it in the next update.

Kapeli:
Should I merge this pull request, or wait?

godbout:
please wait, thanks.

i'm gonna try your fix and see if it helps for the other a-little-off highlighting issue.

godbout:
all good now. can't reproduce the highlighting issue that happened in the v4 of this Docset (even before the tweak mentioned above). weird. but all good. can merge. thanks!

Kapeli:
Thank you! 👍

Kapeli:
> i can make the change, yes. if it's a bug in Dash (it does feel like it), it'd be nice to have it fixed in there too. i don't mind also people having to use the latest version of Dash. i'll update my other Docsets only when i'll get a latest doc. thanks!

I finally got a chance to look into this and I've fixed it. The fix will be out in Dash v5.5.0.

godbout:
excellent. thank you!

PR #3005, opened 2020-10-01
Build Screeps docset

malt3:
I implemented a Docset for the programming MMO screeps. I believe that many users would be interested in using this as playing the game involves reading the  documentation a lot. I am also commited to supply updates as the official documentation is updated.


malt3:
I am not an expert on the tar command but the output may suggest that creating the package on Windows is a problem. I can retry on Linux or macOS.

Kapeli:
You shouldn't use Windows paths in the docset index (docSet.dsidx). Instead of `api\index.html`, use `api/index.html`.

malt3:
I recreated the index using unix paths

Kapeli:
Thank you! 👍

PR #3022, opened 2020-10-16
update TailwindCSS Docset, fix missing Sample entries

godbout:
sorry, some of the doc structure has changed and i missed it so the `Sample` entries disappear. i've added tests on the Docset generation so that it doesn't happen again. the updated Docset here adds back the `Sample` entries. thanks!


Kapeli:
Thank you! 👍

PR #3038, opened 2020-10-31
Add new docset: Pine Script

GeneralD:
I added a new docset about Pine Script.
It is a domain specific language used in TradingView platform.


Kapeli:
The docset's search index (`docSet.dsidx`) is empty.

PR #3067, opened 2020-11-28
Nitrogen docset submission

bunnylushington:
Hi!  I'm submitting a PR for the inclusion of a [Nitrogen](https://nitrogenproject.com) docset.  I **think** I've followed the detailed guidelines but am more than happy to make any adjustments you deem necessary.

There are two small mismatches between Nitrogen terminology and the entry types Dash provides (specifically, "Action" and "Validator" would be handy) but the utility of having the docset available more than makes up for it.

Thanks for your consideration!


Kapeli:
Thank you! 👍

PR #3074, opened 2020-11-30
Octave v6.1.0

obstschale:
- [x] Main Docset replaced
- [x] docset.json updated with new version
- [x] Add version Docset
- [x] Update Readme


Kapeli:
Thank you! 👍

PR #3086, opened 2020-12-09
New docset for PyQGIS

troopa81:
Add PyQGIS API docset for version 3.16.1

QGIS (http://qgis.com/) is an opensource widely used GIS (Geographic Information System) application that could be extended thanks to this API. This docset is related to the Python API (QGIS docset is the C++ one)


Kapeli:
Thank you! 👍

PR #3106, opened 2020-12-25
Update Presto docset to 348

knu:


Kapeli:
The docset's docSet.dsidx seems to be empty (it contains no entries).

Kapeli:
Thank you! 👍

knu:
🙇🏻 They are in the process of renaming their product and I submitted the broken result.  This PR was while I was regenerating all recent releases, so I'll follow up with another PR to include versions up to 350.

PR #3112, opened 2021-01-04
Update Presto docset to 350

knu:
349 was a canceled version.


knu:
Due to a trademark issue raised by Facebook and the new foundation, Presto is announced to be renamed to Trino in a short period of time. https://trino.io/blog/2020/12/27/announcing-trino.html

Is there a recommended way to handle the transition?  Can I just make a new docset with a new CFBundleIdentifier?

Kapeli:
Yes, you can add the docset as a completely new one and users will have to manually choose to install it (they won't get updated from Presto to Trino automatically).

If you want users to update from Presto to Trino automatically, you have to keep the Presto folder name within this repo and update the `docset.json`, docsets and archives to mention Trino instead. That will cause users to only see Trino as the name (no mention of Presto anywhere in Dash), but within this repo the docset will be located within the Presto folder, same as now.

PR #3148, opened 2021-01-31
New Elm packages from 2021-01-31

kraklin:
New docset after such a short time because I'm testing testing the semi-automatic generation of it. And it seems it is working 👍


Kapeli:
Thank you! 👍

PR #3154, opened 2021-02-04
Update to ns-3.33

tnull:
This PR updates the ns-3 docset to version 3.33.

As was the case with the last two updates, the docset size surpasses Github's 100MB limit. 
You can find it [via this Dropbox link](https://www.dropbox.com/s/szuktadtdysfpq2/ns-3.tgz?dl=0).

It would be great if you could place the old version under `versions/3.32/ns-3.tgz`.

Thank you!


Kapeli:
Thank you! 👍

PR #3156, opened 2021-02-05
Add docset for Raspberry Pi Pico SDK

cesarvandevelde:
This is a docset for the C/C++ SDK of the newly released Raspberry Pi Pico. Generator script can be found [here](https://github.com/cesarvandevelde/pico-sdk-docset)


Kapeli:
Docset archive (`Raspberry Pi Pico SDK.tgz`) should not have whitespaces in it. Please replace whitespaces with underscore.

cesarvandevelde:
Updated! The archive filename uses underscores, the uncompressed docset filename still contains spaces. Is this correct?

Kapeli:
Thank you! 👍

PR #3166, opened 2021-02-16
updated OTRS docsets to latest version and added ITSM modules

hanneshal:
Hi 
after a while I updated the OTRS / Znuny (Fork) Docsets for this. 

Regards
Hannes


docsets/OTRS/docset.json
```diff
@@ -12,32 +12,64 @@
...
+        "archive": "versions/3.0/otrs30.tgz"
     },
     {
         "version": "2.4",
-        "archive": "versions/2.4/otrs24.tgz",
+        "archive": "versions/2.4/otrs24.tgz"
     },
+    {
+        "version":"FAQ",
+        "archive": "versions/FAQ.tgz"
```
Kapeli, line 44: This path is wrong. It should be `versions/FAQ/FAQ.tgz`. Paths below are wrong as well.

hanneshal: Thanks! I missed that. It is fixed with the latest commit

Kapeli:
Thank you! 👍

PR #3206, opened 2021-03-14
Swift 5.4 update

roeybiran:


Kapeli:
The docset needs to be archived on its own without any other folder. The archive contains a folder called `.build`.

roeybiran:
fixed

Kapeli:
Thank you! 👍

PR #3217, opened 2021-03-18
MLflow: bump latest version to 1.14.1 and add missing 1.14.0 version to list

imrehg:
Latest one released 2021-03-01.


Kapeli:
Thank you! 👍

PR #3234, opened 2021-03-30
Recreate Standard ML docset with everything else scanned/parsed

toy:
Original version had only top level types scanned so I've tried to properly find types, operators, bindings and hopefully all other stuff which required some preparation of original htmls.


Kapeli:
Two issues:
1. The `docset.json` is missing [`version`](https://github.com/Kapeli/Dash-User-Contributions/blob/master/Sample_Docset/docset.json#L3)
2. Please use "Struct" instead of "Structure" wherever you have it as the type (in `docSet.dsidx` and within page's table of contents)

toy:
1. Could you please suggest what better to use, given that Basis Library has dates in [change history](https://smlfamily.github.io/Basis/history.html)? `"2004.04.12"`?
2. Now I understand why it was displayed as "Variables" in Dash :D

Kapeli:
1. The date itself is fine. You can also have a "hidden version" (one that starts with a `/`). For example, if you set the version as `/2004.04.12`, **no version** will be displayed to the user, but Dash will know when to update the docset.

toy:
Force pushed with both changes applied

Kapeli:
Thank you! 👍

PR #3247, opened 2021-04-10
Add PyMC3 docs

paw-lu:
Closes #1844.
﻿

PR #3267, opened 2021-04-24
Add PyMC3 docs

### LINKED ISSUE #1844: Request: PYMC3 docset
The documentation is available at http://docs.pymc.io/.
Would it be possible to have a dash integration of this documentation?

paw-lu:
Closes #1844.

Resolved icon issue from #3247.


docsets/PyMC3/docset.json
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "PyMC3",
+    "version": "3.11.2",
+    "archive": "Sample.tgz",
```
Kapeli, line 4: I think this is supposed to point to `PyMC3.tgz`.

paw-lu: Good catch!
Addressed in latest commit.

Kapeli:
Thank you! 👍

PR #3325, opened 2021-06-11
added SpigotMC 1.17 javadoc.

vk2gpz:


Kapeli:
Fails validation:

```
Checking SpigotMC/SpigotMC.tgz...
Could not load entries at "SpigotMC-1.17.docset"
Checking SpigotMC/versions/1.17/SpigotMC.tgz...
Could not load entries at "SpigotMC-1.17.docset"
```

As far as I can tell the docSet.dsidx is empty.

vk2gpz:
thanks for checking, i will get that fixed.

PR #3327, opened 2021-06-13
Add WebExtensions docset

private-face:
[WebExtensions API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions) is cross-browser technology for building browser extensions. 

The existing "Chrome Extensions API" docset for Dash, contains only Chrome-specific APIs, while this one covers Firefox (both desktop and mobile), all Chromium-based browsers (Chrome, Brave, Edge, Opera, etc), and Safari.

private.face@gmail.com


Kapeli:
Great job! Thank you! 👍

PR #3340, opened 2021-06-27
Update Docset for AWS CloudFormation Template Reference 2021.06.24

majwhite:
Update Docset for AWS CloudFormation Template Reference 2021.06.24


Kapeli:
Thank you! 👍

PR #3361, opened 2021-07-16
Update to ns-3.34

tnull:
This updates the ns-3 docset to the latest version. As previously, this PR only includes the metadata, the actual docset can be found here: https://www.dropbox.com/s/z2ipcxx3v4sty41/ns-3.tgz?dl=0


Kapeli:
Thank you! 👍

PR #3362, opened 2021-07-18
Update Nomad docsets

blake:
Add recent Nomad docsets, and rebuilds older docsets to fix issues with broken URLs.

Note, this is my first time using `/` to submit updates to existing docset versions, as described in [Docset Versioning Guidelines](https://github.com/Kapeli/Dash-User-Contributions/wiki/Docset-Versioning-Guidelines). I'm not sure if I am using the correct directory layout. I generated a custom [docset feed](https://kapeli.com/docsets#dashdocsetfeed) to test this.

```xml
<!-- Nomad.xml -->
<entry>
  <version>1.1.2</version>
  <url>http://localhost:8000/Nomad.tgz</url>
  <other-versions>
    <version>
      <name>1.1.2/sub_version</name>
    </version>
  </other-versions>
</entry>
```

I observed Dash attempts to download the docsets from the path `/versions/Nomad/1.1.2/sub_version/Nomad.tgz`, whereas the directory hierarchy in this commit is `/versions/1.1.2/sub_version/Nomad.tgz`. I assume the latter is correct because the archive path is explicitly specified as such in docset.json. Please let me know if this is not correct.

### DIFF
--- a/docsets/Nomad/docset.json
+++ b/docsets/Nomad/docset.json
@@ -1,36 +1,166 @@
 {
     "name": "Nomad",
-    "version": "0.12.0",
+    "version": "1.1.2/001-6cba3c5c",
     "archive": "Nomad.tgz",
     "author": {
         "name": "Blake Covarrubias",
         "link": "https://github.com/blake"
     },
-    "aliases": [],
+    "aliases": [
+
+    ],
     "specific_versions": [
         {
+            "version": "1.1.2/001-6cba3c5c",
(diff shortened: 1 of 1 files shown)


Kapeli:
Thank you! 👍 

What this repo does is completely different from docset feeds.

The `/` trick only works for the main docset version (i.e. not the ones under `specific_versions`, which never update). It has no effect on `specific_versions` docsets.

PR #3369, opened 2021-07-24
updated to 9.6.3rc2

jameshale:
some new entries in the dictionary with RC2

### DIFF
--- a/docsets/LiveCode/docset.json
+++ b/docsets/LiveCode/docset.json
@@ -1,6 +1,6 @@
 {
     "name": "LiveCode",
-    "version": "1.9.0",
+    "version": "1.9.1",
     "archive": "LiveCode.tgz",
     "author": {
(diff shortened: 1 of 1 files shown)


docsets/LiveCode/docset.json
```diff
@@ -9,25 +9,21 @@
  
     "specific_versions": [ 
     
+             {
+             "version": "1.9.1",
+             "archive": "versions/1.9.1/LiveCode.tgz"
```
Kapeli, line 14: This archive doesn't exist. I think you added `versions/1.8.1/LiveCode.tgz` instead.

jameshale: Darn. No I will se what I did wrong.
I will hopefully resolve this tomorrow When I look at what I did, and still have.

jameshale:
I will start over.

PR #3392, opened 2021-08-12
DisVoice docset

NeuroForLunch:


docsets/DisVoice/docset.json
```diff
@@ -0,0 +1,12 @@
+{
+    "name": "DisVoice",
+    "version": "0.1.1",
+    "archive": "DisVoice.tgz",
```
Kapeli, line 4: This archive does not exist. You added `disvoice.tgz` instead

Kapeli:
`icon.png` needs to be `16x16` in size and `icon@2x.png` needs to be `32x32`

NeuroForLunch:
fixed

Kapeli:
Thank you! 👍

PR #3395, opened 2021-08-13
Librosa

NeuroForLunch:


docsets/librosa/docset.json
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "librosa",
+    "version": "0.8.1",
+    "archive": "Sample.tgz",
```
Kapeli, line 4: Please change this to `librosa.tgz`

Kapeli:
`icon.png` needs to be `16x16` and `icon@2x.png` needs to be `32x32` in size.

Kapeli:
Why did you change the folder name to `librosa`? Changing the folder name will cause for duplicate entries to be shown to users that already installed the old docset.

docsets/librosa/docset.json
NeuroForLunch: fixed.

NeuroForLunch:
> Why did you change the folder name to `librosa`? Changing the folder name will cause for duplicate entries to be shown to users that already installed the old docset.

The library is actually called librosa or Librosa and not libROSA. I tried to change it back to libROSA on git but it doesn't seem to pick up the capitalization change.

Kapeli:
Use this:

```
git mv librosa temp
git mv temp LibROSA
```

The folder name is used just as the internal identifier in Dash. You can change the name users see in `docset.json`.

Kapeli:
Thank you! 👍

PR #3396, opened 2021-08-13
MNE

NeuroForLunch:
The docsets were too big for github so I uploaded & included the download links for them instead.


Kapeli:
Are all these versions needed or could you skip some of them? As these docsets are rather large, I'd like to keep disk usage to a minimum.

Kapeli:
Also, the floating banner/navigation at the top of pages goes over the anchors. Can you make the banner/navigation static instead of floating?

NeuroForLunch:
Which versions should I keep?

Kapeli:
I'd say just the latest one? Do a lot of people use the older versions?

NeuroForLunch:
I was thinking the last one too

PR #3430, opened 2021-09-22
updated SDL docset to 2.0.16

superzazu:
Hello, here is an update of the SDL docset (2.0.16). It has been a while since I last made a contribution, I hope everything is OK.


Kapeli:
Thank you! 👍

PR #3435, opened 2021-09-27
Add tensorflow_probability v0.14.0

zfc09:


docsets/TensorFlow_Probability/docset.json
```diff
@@ -1,13 +1,17 @@
 {
     "name": "TensorFlow Probability",
-    "version": "0.11.1",
+    "version": "0.14.0",
     "archive": "TensorFlow_Probability.tgz",
```
Kapeli, line 4: You need to update this archive as well.

PR #3517, opened 2021-12-12
Update Hammerspoon docset to 0.9.92

cmsj:


cmsj:
Apologies if this is mangled somehow - I just rewrote our build scripts and maybe missed something.

PR #3519, opened 2021-12-13
adds Python_Sorted_Containers

cad0p:


Kapeli:
`icon.png` needs to be exactly `16x16` in size and you also need to have an `icon@2x.png` which should be `32x32` in size.

PR #3523, opened 2021-12-14
update tailwindcss, fine-tuning UI

godbout:
a thousand apologies. i should have marked the 4 days ago PR in WIP when i started working on the update today.

i'm adding again Alpine here also coz i'm wondering if the one i added today was the last.

both ok through docsetcontrib.

sorry for the extra work!


Kapeli:
Thank you! 👍

PR #3532, opened 2021-12-20
add norns icon to norns docset directory

midouest:
I mistakenly built the icon into the `.docset` package instead of adding it to this directory.

Do I need to update the `version` field or the docset itself for this change to be distributed?


Kapeli:
Yes. You need to update the `version`.

`icon.png` also needs to be exactly `16x16` in size. You also need to add an `icon@2x.png` which needs to be `32x32` in size.

midouest:
Done. Thank you!

Kapeli:
Thank you! 👍

PR #3561, opened 2022-01-20
Add Arm Morello docsets

pthariensflame:
This adds the A64 ISA and System Register docsets for Arm's [Morello](https://www.arm.com/architecture/cpu/morello) platform, supplementing the existing Armv8 and Armv9 docsets.


Kapeli:
Thank you! 👍

PR #3574, opened 2022-02-02
WP-CLI: Update to latest version (2.6) and fill in missing versions

rosswintle:
This docset was way out of date. I've updated it to the latest version and filled in the missing versions.

Hope I did everything right. Let me know if it's not correct.

Props to @schlessera for helping out with the WP-CLI repo today.


Kapeli:
Thank you! 👍

PR #3575, opened 2022-02-03
update tailwind to doc 3.0.18

godbout:
apologies for the quick succession in updates. many changes in 3.0.18 so it took a few days to update to that new doc format.

tested through the docsetcontrib.


Kapeli:
Thank you! 👍

PR #3578, opened 2022-02-05
Add Trino docset version 370

knu:
Trino is formerly known as PrestoSQL.  The Presto docset should probably be updated with that of PrestoDB, but that will be covered by a different PR. (#3579)


Kapeli:
`icon.png` needs to be exactly `16x16` in size and `icon@2x.png` needs to be `32x32` in size.

knu:
@Kapeli OK, fixed!

Kapeli:
Thank you! 👍

PR #3579, opened 2022-02-05
Update Presto docset to 0.269

knu:
Versions from Trino have been removed and will be moved to docsets/Trino. (#3578)


Kapeli:
Thank you! 👍

PR #3599, opened 2022-02-23
Added MOS Protocol 2.8.5 docset.

mbirth:
This is the TCP socket version (2.x) of the MOS Protocol. There are also versions 3.x (HTTP) and 4.x (WSS) of the MOS Protocol in use in parallel. (See http://mosprotocol.com/current-version/ ) That's why I've added the version number to the name here and also tagged it as a specific version.


Kapeli:
The docset folder must not be versioned (i.e. rename `docsets/MOS_Protocol_2.8.5` to `docsets/MOS_Protocol`).

Docsets within `specific_versions` must be in a separate folder than the main docset.

Kapeli:
Thank you! 👍

PR #3644, opened 2022-04-01
Arm A-Profile docsets for update 2022-03

pthariensflame:
Arm once again rearranged and consolidated things so we're doing the same for the docsets. (See leolovenet/Dash-ARMRef#6)


pthariensflame:
Whoops; I'll fix that.

pthariensflame:
Should be ready now.

Kapeli:
Thank you! 👍

Issue #3673, opened 2022-04-27
Terrafrom 1.1.x w/ AWS v4 provider


jwhitcraft:
Hello,

The latest version of Terraform is missing and so is the most recent version of the AWS provider.  I tried running the script that is in the repo, but it's fraught with errors and really outdated.  If someone can provide more detailed instructions on how terrafrom was recently updated, i'd be more than happy to provide an update.

PR #3712, opened 2022-05-27
Terraform 1.2.0.220527

roberth-k:
- Takes over maintenance of the Terraform docset, which hasn't been updated since November.
- Docset is now maintained at https://github.com/roberth-k/dash-docset-terraform.
- Publishes the Terraform docset for Terraform 1.2.0, as well as the latest provider versions as of May 27th.
- Similar to the AWS CLI docset, this will eventually be automatically published on a weekly basis.
- Closes https://github.com/Kapeli/Dash-User-Contributions/issues/3673


Kapeli:
Thank you! 👍

PR #3722, opened 2022-05-28
BungeeCordChat 1.18

shitzuu:


docsets/BungeeCordChat/docset.json
```diff
@@ -1,17 +1,20 @@
...
-    "aliases": ["bungeecordchat"
-    ],
+    "aliases": [ "bungeecordchat" ],
     "specific_versions": [
         {
             "version": "1.15",
             "archive": "versions/1.15/BungeeCordChat.tgz"
+        },
+        {
+            "version": "1.18",
```
Kapeli, line 16: Newer versions should be ordered at the top. I'll fix it.

PR #3755, opened 2022-06-21
Paper 1.19 - updating the corrupted docset

shitzuu:
Replaced the docsets deployed in the previous pull request #3753, as they were missing methods defined in interfaces.


Kapeli:
Thank you! ❤️

PR #3764, opened 2022-06-29
Update RuboCop docset to v1.30.1 and v1.31.1

milouse:


milouse:
Woops, I’ll take care of that.

PR #3787, opened 2022-07-13
update godot docset to the latest stable (3.4.4)

poga:
The actual [`godot.tgz`] can be downloaded from [google drive](https://drive.google.com/file/d/10L5PqqMMvD5rtlfHDbfCRIFDPpF8gqJ9/view?usp=sharing) since it's larger than 100MB(186MB).

The docset is generated and optimized with [a custom script](https://github.com/poga/godot-dash-docset).

Please let me know if there's any issue when downloading :)

## checklist

- [x] You must maintain your docset! Always update it to the latest stable version as soon as possible
- [x] You must try to fix any issues/bugs that might be reported

- [x] The docset name must match the official API name (e.g., if you make a docset for PostgreSQL, don't call it psql)
- [x] The docset must not include a version in its Info.plist or in the docset filename
- [x] All pages must work offline (online resources are okay only if there's absolutely no way to avoid them)
- [x] The docset index must not contain entries with new lines in them or empty entries or broken paths
- [x] The docset must be for public use (i.e. don't contribute a docset that only you or your team will use)

- [x] Docset should have an icon
- [x] Docset should have an [index page set](http://kapeli.com/docsets#settingindexpage)
- [x] Docset should have [JavaScript enabled](http://kapeli.com/docsets#enableJavascript) if it needs it
- [x] Docset should have [table of contents support](http://kapeli.com/dash_guide#tableOfContents) if it makes sense
- [x] Docset should be optimised for display in Dash

### DIFF
--- a/docsets/Godot/README.md
+++ b/docsets/Godot/README.md
@@ -3,52 +3,17 @@ Godot Docset
 
 [Godot](https://godotengine.org/) docset for [Dash](http://kapeli.com/dash).
 
-By [Dmitrii Maganov](https://github.com/vonagam).
+Generated and optimized by [poga](https://github.com/poga)
 
-Instructions how to reproduce the build (using version 3.2 as an example):
+## Build
 
-- Install [Dashing](https://github.com/technosophos/dashing) (`brew install dashing` on mac).
+The docset is generated with a custom script to optimize reading experience in dash.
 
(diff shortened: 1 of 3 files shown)


Kapeli:
Thank you! ❤️

PR #3805, opened 2022-07-24
update scikit-learn to 1.1.1 and the build script.

tshu-w:


tshu-w:
I'm trying to use official docs from https://scikit-learn.org/dev/versions.html directly. I can only find `genindex` and `py-modindex` in `_static/basic.css` and I don't know why this test failed.
```css
div.modindex-jumpbox {
    border-top: 1px solid #ddd;
    border-bottom: 1px solid #ddd;
    margin: 1em 0 1em 0;
    padding: 0.4em;
}
div.genindex-jumpbox {
    border-top: 1px solid #ddd;
    border-bottom: 1px solid #ddd;
    margin: 1em 0 1em 0;
    padding: 0.4em;
}
table.modindextable td {
    padding: 2px;
    border-collapse: collapse;
}
```

Kapeli:
You have entries in docSet.dsidx which point to files that don't exist. For example, entry with row id "3058" has the name "Index" and the path "genindex.html", but "genindex.html" does not exist.

PR #3812, opened 2022-07-31
scikit: update to 1.1.1

tshu-w:


tshu-w:
Building from source also has this problem and may require modifying the parser, which is a bit out of my scope.

Kapeli:
`doc2dash` is not maintained by me. If you think this is an issue with `doc2dash`, please report it there.

PR #3842, opened 2022-08-21
Update Neovim docset to 0.7.2

h-hg:
The Neovim docs has not been updated for a long time, and the original maintainer seems to be no longer maintaining it.


Kapeli:
Thank you! ❤️ 

Note: you need to update the main archive (located at `docsets/Neovim/Neovim.tgz`) too. I'll update it by copying `docsets/Neovim/versions/0.7.2/Neovim.tgz`.

PR #3850, opened 2022-08-24
Updated to include LC10DP4 dictionary changes

jameshale:
DP4 modified some dictionary entries updating them and also added further examples of syntax.
Some new items also added.


Kapeli:
Thank you! ❤️ 

Note: newer versions should be ordered at the top in `specific_versions`. I'll change the order.

PR #3856, opened 2022-08-28
FEATURE: add guzzle docset

crydotsnake:
Hello!

Right now i dont have an icon. But i contacted the developers of guzzle already. If i have the icons, i will create a new pull request.


Kapeli:
If you don't have an icon please don't include one.

Failed validation: `File not found at "py-modindex.html" for entry with row id "12" of "guzzle.docset"`. This mean that the entry with row id 12 in the docSet.dsidx points to a file that doesn't exist (case-sensitive).

crydotsnake:
I dont know why this always happens...

PR #3899, opened 2022-09-22
Update more-itertools docset 4.3.1 -> 8.13.0

gkoller:
Updates the `more-itertools` docset from version 4.3.1 to 8.13.0. While doing so:

- renamed the docset from `more_itertools` to `more-itertools`. The latter is the actual project name. Retained old docset name as an alias
- changed docset generation script from Dashing to [doc2dash](https://github.com/hynek/doc2dash)
- retained the original documentation's layout and formatting (Sphinx RTD theme) minus the navigation sidebar (old docset choose entirely different theme)


Kapeli:
Great job!

You need to keep using the old folder (`more_itertools`) in order for users that already have the old `more_itertools` docset installed to get updated to your new docset. Dash uses the folder name as the unique identifier of a docset, so if the folder name is changed then the docset is treated like a completely new one and that causes old users to not get the updated docset.

Please rename the folder back to `more_itertools`. This folder name is not displayed to users anywhere, all mentions of the docset will become `more-itertools` as far as users are concerned.

Kapeli:
Thank you! ❤️

gkoller:
> Please rename the folder back to `more_itertools`. 

Done. And thank you for the explanation why it should remain so.

PR #3920, opened 2022-10-11
Update Neovim to 0.8.0

h-hg:


docsets/Neovim/docset.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "Neovim",
-    "version": "0.7.2",
+    "version": "0.8.0",
     "archive": "Neovim.tgz",
```
Kapeli, line 4: You need to update this archive as well. I'll update it by copying the one at `versions/0.8.0/Neovim.tgz`. If that's not ok, please let me know.

PR #3931, opened 2022-10-16
Update Wagtail docset to version 2.12

activus-d:


Kapeli:
Please submit a single pull request that contains all versions.

activus-d:
@Kapeli Sorry to bother you but this is my first attempt at trying this. My question is how that since we have to leave a duplicate archive file of each version in the `docset/wagtail` folder, using same name format would make any subsequent archive file overwrite the preceding one. Should we compromise on the name format of each archive file`(For instance “wagtail.tgz” should be renamed as “wagtail-2.9.tgz”)` or allow the subsequent file to overwrite the preceding one.

PR #3960, opened 2022-10-27
Update Trino docset to 401

knu:
- Fix anchors to types
- Support new, improved section structures and markups
- Add general properties as variables


Kapeli:
Thank you! ❤️

PR #3961, opened 2022-10-28
juce 7.0.2

modosc:
add `juce-7.0.2`

### DIFF
--- a/docsets/JUCE/docset.json
+++ b/docsets/JUCE/docset.json
@@ -1,6 +1,6 @@
 {
     "name": "JUCE",
-    "version": "6.1.6",
+    "version": "7.0.2",
     "archive": "JUCE.tgz",
     "author": {
         "name": "modosc",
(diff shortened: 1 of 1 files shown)


docsets/JUCE/docset.json
```diff
@@ -11,6 +11,10 @@
         "Jules Utility Class Extensions"
     ],
     "specific_versions": [
+        {
+            "version": "7.0.2",
+            "archive": "versions/7.0.2/JUCE.tgz"
```
Kapeli, line 16: This file does not exist. You added `versions/7.1.2/JUCE.tgz` instead

modosc: ack! it was supposed to be 7.0.2 but i typo'd the branch name and used that. should be fixed now?

Kapeli:
Thank you! ❤️

PR #3974, opened 2022-11-07
feat: update three.js r146

### LINKED ISSUE #3973: update Threejs docset to r146

mzvast:
close #3973


docsets/Three.js/docset.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "Three.js",
-    "version": "r77",
+    "version": "r146",
     "archive": "threejs.tgz",
```
Kapeli, line 4: Please note that you should update this archive as well. I'll update it by copying `versions/r146/threejs.tgz`. If this is not ok, please let me know.

mzvast: It's correct,thank you:)

PR #3991, opened 2022-11-18
Update BigQuery Standard SQL docset to 2022.11.18

knu:
I see some updates in features regarding external tables.


Kapeli:
Thank you! ❤️

PR #3993, opened 2022-11-19
Add docset for nox 2022.8.7

paw-lu:
Add docset for nox 2022.8.7.

This pull request was created by [paw-lu/nox-dash-docset](https://github.com/paw-lu/nox-dash-docset),


Kapeli:
Thank you! ❤️

PR #3998, opened 2022-11-21
Add docset for invoke 1.7.3

paw-lu:
Add docset for invoke 1.7.3.

This pull request was created by [paw-lu/invoke-dash-docset](https://github.com/paw-lu/invoke-dash-docset),


Kapeli:
Thank you! ❤️

PR #3999, opened 2022-11-22
Add docset for polars 0.14.29

paw-lu:
Add docset for polars 0.14.29.

This pull request was created by [paw-lu/polars-dash-docset](https://github.com/paw-lu/polars-dash-docset),


Kapeli:
Thank you! ❤️

PR #4002, opened 2022-11-23
Add docset for pymc 4.4.0

paw-lu:
Add docset for pymc 4.4.0.

This pull request was created by [paw-lu/pymc-dash-docset](https://github.com/paw-lu/pymc-dash-docset),


Kapeli:
Thank you! ❤️

Issue #2601, opened 2019-12-18
Request: AWS CDK (Cloud Development Kit)


baxang:
https://github.com/aws/aws-cdk
https://docs.aws.amazon.com/cdk/api/latest/

The CDK is available in the following languages:

- JavaScript, TypeScript (Node.js ≥ 10.3.0)
- Python (Python ≥ 3.6)
- Java (Java ≥ 8 and Maven ≥ 3.5.4)
- .NET (.NET Core ≥ 3.0)

nmyster:
Any updates on this or how we get it

likeabbas:
@huyphan has created the script to make the Typescript docset here https://github.com/huyphan/aws-cdk-typescript-dash-docset

Now we just need him to complete the PR for it.

dannysteenman:
I noticed there was also an open issue on the aws-cdk repo: https://github.com/aws/aws-cdk/issues/8617. I mentioned in that issue that someone already did the heavy lifting in https://github.com/huyphan/aws-cdk-typescript-dash-docset/issues/1

Maybe both parties can collaborate in order to get this into Dash 👍

PR #4026, opened 2022-12-01
AWS CDK docset

roberth-k:
- Based on https://docs.aws.amazon.com/cdk/api/v2/docs/aws-construct-library.html.
- Docset maintained in https://github.com/roberth-k/dash-docset-aws-cdk.
- The initial version is `2.51.1/1669916658`. This includes the hidden version segment documented in https://github.com/Kapeli/Dash-User-Contributions/wiki/Docset-Versioning-Guidelines.
- A new docset will automatically be published for every new version of the CDK.
- Closes https://github.com/Kapeli/Dash-User-Contributions/issues/2601.


Kapeli:
Thank you! ❤️

PR #4028, opened 2022-12-02
Add GNU Autoconf 2.71

lshprung:
Adds GNU Autoconf docset (sourced from [here](https://www.gnu.org/software/autoconf/manual/)). This docset complements the existing GNU Automake docset.


Kapeli:
Thank you! ❤️

PR #4031, opened 2022-12-03
Update Vulkan docset to version 1.3.236

antiagainst:


docsets/Vulkan/docset.json
```diff
@@ -15,6 +15,10 @@
         "API"
     ],
     "specific_versions": [
+        {
+            "version": "1.3.236",
+            "archive": "versions/1.3.236/Vulkan.tgz"
```
Kapeli, line 20: This file doesn't exist. I think the file was added at this path instead: `docsets/Vulkan/versions/1.3.236`.

antiagainst: Fixed. Thanks!

PR #4049, opened 2022-12-12
updated SDL docset to 2.26.1

superzazu:
Hello, here is an update of the SDL docset (2.26.1). Please let me know if I made a mistake.


Kapeli:
Thank you! ❤️

PR #4058, opened 2022-12-19
Added debmake 1.17

lshprung:
debmake documentation as sourced from [debmake-doc](https://tracker.debian.org/pkg/debmake-doc) and compiled using tools [here](https://github.com/lshprung/debmake-dash-docset). The docset included in this PR is in English, but it is possible to compile in several different languages. This docset provides guides for building .deb packages


Kapeli:
Thank you! ❤️

PR #4070, opened 2022-12-25
add OKD docset (from v4.4 to v4.11)

anisimovdk:


docsets/OKD/docset.json
```diff
@@ -0,0 +1,42 @@
+{
+    "name": "OKD",
+    "version": "4.11",
+    "archive": "OKD.tgz",
```
Kapeli, line 4: File paths are case-sensitive. You need to rename the archives to `OKD.tgz`.

anisimovdk: fixed

Kapeli:
Thank you! ❤️

PR #4078, opened 2022-12-30
Update Presto docset to 0.278.1

knu:
This includes all versions after the last update.


Kapeli:
Thank you! ❤️

knu:
Ah, previous versions are lost in the last push.  I'll follow up later.

PR #4079, opened 2022-12-30
Added GNU Autoconf Archive 2022.09.03

lshprung:
GNU Autoconf Archive is a set of m4 macros for use in GNU Autoconf projects. As such, this docset should compliment the existing GNU Autoconf docset.


Kapeli:
Thank you! ❤️

PR #4080, opened 2022-12-30
ncurses 6.4

lshprung:
Bump ncurses docset to version 6.3. Please note that the docset generation method was [redone](https://github.com/lshprung/ncurses-dash-docset), and documentation is now sourced from [upstream](https://invisible-island.net/ncurses/) rather than from Debian man pages.


lshprung:
Packaged the just released version 6.4 over 6.3.

PR #4084, opened 2022-12-31
Added GNU Libtool 2.4.7

lshprung:
Adds GNU Libtool docset (sourced from [here](https://www.gnu.org/software/libtool/manual/)). the GNU Libtool docset, along with the GNU Autoconf and GNU Automake docsets, provide dash docsets for all of the major components of the GNU Autotools.


Kapeli:
Thank you! ❤️

PR #4087, opened 2023-01-02
Add Remix docset

gazoakley:
Hi - this is a first contributed docset for me, so feedback/comments appreciated


Remix/docset.json
```diff
@@ -0,0 +1,14 @@
+{
+    "name": "Remix",
+    "version": "1.9.0",
+    "archive": "remix.tgz",
```
Kapeli, line 4: This file doesn't exist. You added `Remix.tgz` instead (note: case-sensitive).

Kapeli:
The white icon isn't really a good idea. It won't look ok in Dash.

Remix/docset.json
gazoakley: Fixed - and the docset isn't in the `docsets` directory either 👍

Kapeli:
Thank you! ❤️

PR #4089, opened 2023-01-02
Added GNU Coding Standards 2021.07.01

lshprung:
Adds GNU Coding Standards docset (sourced from [here](https://www.gnu.org/prep/standards/)). I went with "gnustandards" for the DocSetPlatformFamily, but I wasn't sure if something else might be better. Let me know if you have any feedback.


Kapeli:
Thank you! ❤️

PR #4097, opened 2023-01-06
Update boto3 to 1.26.44

jamesoff:
I added my Mastodon profile, as I'm using that more than Twitter now. Hope that's ok; happy to revert to just Twitter if that's preferred.


jamesoff:
Is that warning a problem? I just had a poke around the live docs (https://boto3.amazonaws.com/v1/documentation/api/latest/index.html) and that `shortbread` script/request is present on all of them - it's in the footer.

Kapeli:
> Is that warning a problem? 

It depends. Some online requests delay the page load and when users have a poor internet connection it might cause issues. Many user-contributed docsets make online requests and even some of the docsets I maintain do it.

PR #4110, opened 2023-01-17
Fix guide image regression

jameshale:
location of guide image files had been changed resulting in them being omitted, from docset.
Now corrected.


docsets/LiveCode/docset.json
```diff
@@ -16,16 +20,16 @@
             "archive": "versions/1.8.0/LiveCode.tgz"
         },
         {
-            "version": "1.7.7",
```
Kapeli, line 19: Newer versions should be ordered at the top.

PR #4131, opened 2023-01-28
Update PEPs docset and build with Sphinx and local links

warrenalphonso:
The current PEP docset is scraped from the [Python PEPs website](https://peps.python.org/). I've built the documentation with Sphinx and converted it with `doc2dash` to generate this Docset.


Kapeli:
Thank you! ❤️

PR #4153, opened 2023-02-11
juce 7.0.5

modosc:
`juce-7.0.5`

### DIFF
--- a/docsets/JUCE/docset.json
+++ b/docsets/JUCE/docset.json
@@ -1,6 +1,6 @@
 {
     "name": "JUCE",
-    "version": "7.0.4",
+    "version": "7.0.5",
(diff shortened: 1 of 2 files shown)


docsets/JUCE/docset.json
```diff
@@ -12,7 +12,7 @@
     ],
     "specific_versions": [
         {
-            "version": "7.0.4",
+            "version": "7.0.5",
             "archive": "versions/7.0.4/JUCE.tgz"
```
Kapeli, line 16: You need to change this path too I think.

modosc: good catch.

Kapeli:
Thank you! ❤️

PR #4176, opened 2023-02-26
Add docset for arviz 0.15.0

paw-lu:
Add docset for arviz 0.15.0.

This pull request was created by [paw-lu/arviz-dash-docset](https://github.com/paw-lu/arviz-dash-docset),


Kapeli:
Thank you! ❤️

PR #4210, opened 2023-03-20
GNU Guile 3.0.9

### LINKED ISSUE #4202: Separate docsets for GNU Guile 2 and 3
Hello. I see that the GNU Guile docset is still at version 2.2. I would like to see version 3 packaged, but recognize it may be beneficial to have docsets available for both GNU Guile 2 and 3. I'm not sure if the best way of accomplishing this would be to have separate docsets or if there is a way to make both versions available via the `versions` directory. Once the best way is determined, I will be happy to create a PR.

lshprung:
Version 2.2.1 was copied as a specific version, as advised in #4202.

Closes #4202


Kapeli:
Thank you! ❤️

PR #4239, opened 2023-04-05
IntelASMx86 2022-09-15 update

pthariensflame:


pthariensflame:
I'm not sure how to solve these errors, since I'm just generating based on an existing script.

Kapeli:
Thank you! ❤️

PR #4246, opened 2023-04-09
AWS-CDK-Python v2.73.0-1681410656

yo-ga:
- Document: AWS CDK Python
- Version: v2.73.0-1681410656
- Generated from: [yo-ga/dash-docset-aws-cdk-python](https://github.com/yo-ga/dash-docset-aws-cdk-python/)


Kapeli:
Thank you! ❤️

PR #4267, opened 2023-04-23
Adds Docassemble docset

thedeletosh:
Docset for Docassemble, a free, open-source expert system for guided interviews and document assembly, based on Python, YAML, and Markdown.


Kapeli:
Please don't delete the sample docset.

`icon.png` needs to be exactly 16x16 in size and you also need to add `icon@2x.png` which should be 32x32 in size.

thedeletosh:
@Kapeli done.

Kapeli:
Thank you! ❤️

PR #4269, opened 2023-04-24
Add docset for nox 2022.11.21

paw-lu:
Add docset for nox 2022.11.21.

This pull request was created by [paw-lu/nox-dash-docset](https://github.com/paw-lu/nox-dash-docset),


Kapeli:
Thank you! ❤️

PR #4276, opened 2023-04-27
Update boto3 to 1.26.121

jamesoff:
As before, the tgz file is now too big to push directly to github.

The file is available to download from https://d24v0t7zqzj9pw.cloudfront.net/1.26.121/boto3.tgz - this link will function for 90d.

Thanks!


Kapeli:
Thank you! ❤️

PR #4278, opened 2023-04-28
Update Vulkan docs to 1.3.249

pthariensflame:


Kapeli:
Thank you! ❤️

antiagainst:
Thanks for the update! However, this changes the original doc style entirely---now all the spec is without the original vulkan spec theming and colors. E.g.:

Before: 
Screenshot 2023-05-02 at 4 45 19 PM

vs. after:
Screenshot 2023-05-02 at 4 45 40 PM

I'm not sure this is the way we want to go?

pthariensflame:
Oops!

Feel free to send in a correct version, and maybe document how to do it?  I was attempting to go by the README, but it was very obviously out of date.

antiagainst:
No worries! I've uploaded https://github.com/Kapeli/Dash-User-Contributions/pull/4290 for version 1.3.250 to fix this. :)

Yes good points. I'll update the README to reflect the current status.

PR #4325, opened 2023-05-31
fix archive error; update docset.json

scillidan:
My fault, I found an extra duplicate folder in the  archive.


Kapeli:
Thank you! ❤️

PR #4421, opened 2023-08-09
Add Golly 4.2

Mick235711:
Docset for [Golly](https://golly.sourceforge.io/), an open-source, cross-platform application for exploring Conway's Game of Life and many other types of cellular automata.


Mick235711:
That file only exists online, weird. I'll add it manually then.

PR #4494, opened 2023-08-17
Update boto3 to 1.28.28

jamesoff:
As usual, the tar file is available for download at https://d24v0t7zqzj9pw.cloudfront.net/1.28.28/boto3.tgz for 90d.

Thanks!


Kapeli:
Thank you! ❤️

PR #4558, opened 2023-09-01
Fix Nim v2.0.0

Archargelod:
Previous docset was incomplete (only modules and sections were indexed).
- Updated icon.
- Updated url to the latest version of script.


Kapeli:
Thank you! ❤️

PR #4595, opened 2023-09-26
Bump pyro

ivaquero:


Kapeli:
Dash uses the folder name as the unique identifier for the docset, so if you change the folder name from `pyro` to `Pyro`, everyone that has `pyro` installed will not get the update.

Please keep the folder name as `pyro`, and change the docset name within `docset.json` and the actual docset.

ivaquero:
Alright, let me recover it.

docsets/pyro/docset.json
```diff
@@ -1,18 +1,24 @@
 {
     "name": "pyro",
```
Kapeli, line 2: You can change the name to `Pyro` here, if you want.

ivaquero: I'm okay with both.

Kapeli:
Thank you!

PR #4597, opened 2023-09-26
bump NetworkX

ivaquero:


ivaquero:
NetworkX document has some problems, closed for the moment.

PR #4628, opened 2023-10-06
add Faker.py

ivaquero:


docsets/Faker.py/docset.json
```diff
@@ -0,0 +1,16 @@
...
+  "archive": "Faker.tgz",
+  "author": {
+    "name": "Xavier Yang",
+    "link": "https://github.com/ivaquero"
+  },
+  "aliases": ["Faker", "Python", "Python Faker"],
+  "specific_versions": [
+    {
+      "version": "19.6.2",
+      "archive": "versions/Faker.tgz"
```
Kapeli, line 13: This path doesn't exist.

ivaquero: Oh, that's a mistake. I will solve it later.

Kapeli:
Thank you!

PR #4633, opened 2023-10-09
Update Tailwind CSS docset to 3.3.3-1

knu:
- Take account of the padding and adjust the scroll gap
- Find more classes and modifiers from outside of tables


Kapeli:
Thank you!

PR #4650, opened 2023-10-19
Update TLDR pages docset to 0.0.3.7 and change author

cvn:
Hello,

This update includes a change to the author information. The original author has been unresponsive for some time. I have been maintaining this docset, and I think it's worth making it official.

Thanks,
Chad


Kapeli:
Thank you!

PR #4660, opened 2023-10-22
update pymc (correct last version)

ivaquero:


Kapeli:
You added `pymc.tgz`, but `docset.json` points to `PyMC.tgz`.

Kapeli:
Thank you!

PR #4725, opened 2023-12-07
Add sqlite-utils docset

Florents-Tselai:
This adds a docset for [sqlite-utils](https://github.com/simonw/sqlite-utils)

It also includes a `Makefile`, which further automates the process for most `sphinx`-powered documentation packages.


Florents-Tselai:
Oops. Forgot to actually archive the docsets. 
Updating the pr...

Florents-Tselai:
Looks like the CI doesn't properly handle multiple versions at once. Trying only the latest.

Florents-Tselai:
If this is accepted, then I'd strongly suggest extracting the Makefile into a common directory and have it include-able in future docsets. ( I can open a separate PR for that)

docsets/sqlite-utils/docset.json
```diff
@@ -0,0 +1,21 @@
...
+  "aliases": [
+  ],
+  "specific_versions": [
+    {
+      "author": {
+        "name": "Florents Tselai",
+        "link": "https://github.com/Florents-Tselai"
+      },
+      "version": "3.35.2",
+      "archive": "versions/3.35.2/sqlite-utils.tgz"
```
Kapeli, line 18: This archive needs to be added as well.

Kapeli:
> If this is accepted, then I'd strongly suggest extracting the Makefile into a common directory and have it include-able in future docsets. ( I can open a separate PR for that)

Dash is now able to generate docsets for Sphinx docs on its own now, starting with version 7.2.0.

docsets/sqlite-utils/docset.json
Florents-Tselai: Added.

Kapeli:
Thank you!

PR #4739, opened 2023-12-16
Xojo 2023r4 Docset

patricksalo:
Addition of the Xojo 2023r4 Docset


docsets/Xojo/docset.json
```diff
@@ -0,0 +1,21 @@
+{
+    "name": "Xojo 2023r4",
```
Kapeli, line 2: The name of the docset should not include the version

docsets/Xojo/docset.json
```diff
@@ -0,0 +1,21 @@
...
+    "version": "1.0.0",
```
Kapeli, line 3: This should be the version of the framework (i.e. use 2023r4 here). See https://github.com/Kapeli/Dash-User-Contributions/wiki/Docset-Versioning-Guidelines

docsets/Xojo/docset.json
```diff
@@ -0,0 +1,21 @@
...
+    "archive": "Xojo_2023r4.tgz",
```
Kapeli, line 4: Don't include the version in the archive name.

Kapeli:
Please remove the "- Xojo Documentation" and "copy to clipboard" from docset entries.

docsets/Xojo/docset.json
patricksalo: I'm happy to change this to just Xojo, but alas it's an interesting beast in that the documentation is tied to the actual release version of the product. So although I can change to just Xojo, the Docset only applies to the 2023 release 4 version of the product. See my next comment below.

docsets/Xojo/docset.json
patricksalo: Continuing on with the prior comment train of thought, there needs to be some way of versioning the Docset. If I replace 1.0.0 with 2023r4, then how do you version the Docset when bugs are found or there are revisions? I'm happy to adjust things to your liking but just wanted to call out that Xojo's releases and documentation is actually quite different than much of the rest of the industry.

docsets/Xojo/docset.json
patricksalo: See my comments above and let me know how you'd like to proceed.

patricksalo:
> Please remove the "- Xojo Documentation" and "copy to clipboard" from docset entries.
> 
> The docset archive should include just the docset. Currently, it includes some parent folders (Users/patrick/Desktop).

Done and just awaiting resolutions to the version comments above.

docsets/Xojo/docset.json
Kapeli: From https://github.com/Kapeli/Dash-User-Contributions/wiki/Docset-Versioning-Guidelines:

* You should use the same version as the API the docset is for
* Anything that comes after a `/` inside the version is not displayed to the user at all. You can submit updates without changing the displayed version by adding anything after a `/`. For example, if the version you want to display to the user is `1.0.0`, you can still make Dash update a docset by setting the version to `1.0.0/whatever` and the user will only see `1.0.0`

So basically you'd set the version to `2023r4` (the version users see) and if you ever want to release an update you can add your own versioning after a `/`, e.g. `2023r4/whatever`

docsets/Xojo/docset.json
Kapeli: Docset versioning is handled by the `version` and not the docset name/other places.

patricksalo:
Updates processed per your requests Bogdan.

Kapeli:
Failed validation:

```
Xojo/Xojo.tgz: FAIL ❌
Empty path found for row id "3611" at "Xojo.docset"
```

patricksalo:
I just noticed the "Failed validation" error above. Not really sure what this means, but I re-TAR'd the docset using a different method to see if that helps. The download link is the same and hasn't changed.

Kapeli:
The failed validation is because you have an entry within docSet.dsidx which does not have a path set.

patricksalo:
> The failed validation is because you have an entry within docSet.dsidx which does not have a path set.

Not really sure what to do here as the docSet.dsidx was generated via the Docset Generator within Dash.

**Thoughts on how to proceed?**

Kapeli:
Sounds like a bug in the Docset Generator. Can you please export the docset config and send it to me?

patricksalo:
> Sounds like a bug in the Docset Generator. Can you please export the docset config and send it to me?

Here you go...

[Xojo 2023r4.docsetconfig.zip](https://github.com/Kapeli/Dash-User-Contributions/files/13893865/Xojo.2023r4.docsetconfig.zip)

patricksalo:
> This docset config indexes local files. Can you please also send the files you're indexing?

[Sure, here you go...](https://www.dropbox.com/scl/fi/54kf9cs8ogpkvyoopnizx/Docs2023r4.tgz?rlkey=pp9mi98hr5n10tohx9f6tb4kn&dl=0)

Kapeli:
I've fixed it. Please re-run the docset generator from Dash and resubmit the docset.

patricksalo:
Okay done and the originally provided checking [download URL](https://www.dropbox.com/scl/fi/cr8axn7ufetdjz95za8ib/Xojo.tgz?rlkey=r4ay8agwhimto4tj6khu9sc0m&dl=0) remains the same...

patricksalo:
Woohoo! Many thanks Bogdan. :)

PR #4779, opened 2024-01-17
Update Presto docset to 0.285

knu:
I'm not sure what would be a good entry type for REST API endpoints like "GET /v1/query".  They are currently indexed as functions.


Kapeli:
"Request" would be an option.

PR #4836, opened 2024-02-26
renpy-engine-8.2.0; update README.md

scillidan:


docsets/RenPy_Engine/docset.json
```diff
@@ -1,10 +1,16 @@
 {
     "name": "RenPy_Engine",
-    "version": "8.1.3",
+    "version": "8.2.0",
     "archive": "RenPy_Engine.tgz",
```
Kapeli, line 4: You need to update the main archive as well, not just the specific version one.

scillidan: I put the archive in `. / versions/8.2.0`. I checked the doc's version. It is indeed `8.2.0`. But there is only one version, should I move it to `./`.

Kapeli: You need to have the archive at these 2 locations:
```
docsets/RenPy_Engine/RenPy_Engine.tgz
docsets/RenPy_Engine/versions/8.2.0/RenPy_Engine.tgz
```

Specific version archives are completely separate from the main version archive.

PR #4884, opened 2024-03-28
Update BigQuery Standard SQL docset to 2024.03.28

knu:
`CREATE AGGREGATE FUNCTION` just added in the last update is gone.


Kapeli:
Thank you!

PR #4896, opened 2024-04-03
Update to Xojo 2024r1 DocSet

patricksalo:
This update to the Xojo 2024r1 docket.json now includes the prior Xojo 2023r4 version. Additionally I adjusted the aliased to make more sense.

Note: Under /versions I've segmented each version into its own folder with a "Xojo Download Link.txt" that will download the proper file.


patricksalo:
Many thanks!

PR #4908, opened 2024-04-08
mypy-1.9 and mypy-1.8

timo:


docsets/mypy/docset.json
```diff
@@ -1,12 +1,25 @@
 {
     "name": "mypy",
-    "version": "0.9.1",
-    "archive": "mypy.tgz",
+    "version": "1.9.0",
+    "archive": "versions/mypy.1.9.tgz",
```
Kapeli, line 4: You shouldn't use versions for this path. The main docset archive should be outside the versions folder.

docsets/mypy/docset.json
```diff
@@ -1,12 +1,25 @@
...
 
-    "specific_versions": []
+    "specific_versions": [
+        {
+            "version": "1.8.0",
+            "archive": "versions/mypy.1.8.tgz"
+        },
+        {
+            "version": "0.9.1",
+            "archive": "mypy.tgz",
```
Kapeli, line 18: This path should be inside the `versions` folder.

PR #4937, opened 2024-05-02
update pymc to 5.14.0

ivaquero:


docsets/pymc/docset.json
```diff
@@ -14,6 +14,10 @@
     {
       "version": "5.9.0",
       "archive": "versions/5.9.0/PyMC.tgz"
+    },
+    {
+      "version": "5.14.0",
+      "archive": "versions/5.14.0/PyMC.tgz"
```
Kapeli, line 20: This file does not exist (paths are case-sensitive).

docsets/pymc/docset.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "PyMC",
-  "version": "5.9.0",
+  "version": "5.14.0",
   "archive": "PyMC.tgz",
```
Kapeli, line 4: You need to update this file as well.

Kapeli: You need to update this file as well. The main version archives is separate from the specific versions.

docsets/pymc/docset.json
```diff
@@ -14,6 +14,10 @@
     {
       "version": "5.9.0",
       "archive": "versions/5.9.0/PyMC.tgz"
+    },
+    {
+      "version": "5.14.0",
```
Kapeli, line 19: Newer versions should be sorted at the top.

Kapeli:
The [main version archive](https://github.com/Kapeli/Dash-User-Contributions/pull/4937/files#diff-37bd9806b755942cd3c43f95ea49e86334ee6663e76c20b428946a72a04854c9R4) is separate from specific version archives. Please update that archive as well.

Kapeli:
Thank you!

PR #4955, opened 2024-05-12
Update Vaadin to version 8.25.1

ewall:
New minor version update...

### DIFF
--- a/docsets/Vaadin/docset.json
+++ b/docsets/Vaadin/docset.json
@@ -1,6 +1,6 @@
-{
+ate{
     "name": "Vaadin",
-    "version": "8.20.3",
+    "version": "8.25.1",
     "archive": "Vaadin.tgz",
     "author": {
         "name": "Eric Wallace",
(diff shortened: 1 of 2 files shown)


docsets/Vaadin/docset.json
```diff
@@ -1,6 +1,6 @@
-{
+ate{
```
Kapeli, line 1: Typo?

docsets/Vaadin/docset.json
```diff
@@ -11,6 +11,10 @@
         {
             "_comment": ""
         },
+        {
+            "version": "8.25.1",
+            "archive": "versions/8.25.1/Vaadin.tgz"
```
Kapeli, line 16: This file doesn't exist. It was added at `docsets/Vaadin/versions/Vaadin.tgz` instead I think.

docsets/Vaadin/docset.json
ewall: There were some weird errors in the doc conversion, but oddly it still worked okay for me?

In any case, we have a new version now so I'll close out this one.

PR #5026, opened 2024-07-01
Zig v0.13.0

kassane:
rel notes: https://ziglang.org/download/0.13.0/release-notes.html


Kapeli:
Thank you!

ldelossa:
@kassane 

Does this docset actually include std lib docs? I load it into dash and I only find the Zig guide, builtins, and keywords.

kassane:
> Does this docset actually include std lib docs? I load it into dash and I only find the Zig guide, builtins, and keywords. 

Try use https://github.com/puqeko/zig2dash

Zig 0.13.0 has changes in stdlib doc.

cc: @puqeko

ldelossa:
@kassane thanks, I'll give this a go.

ldelossa:
Yup it is indeed an issue, see: https://github.com/puqeko/zig2dash/issues/1

puqeko:
Unfortunately this may be because of fundamental changes to how the zig docs are served which will mean another approach to generating static docs is needed

puqeko:
It seems that since Zig 0.12.0 you can generate [fully interactive offline documentation with the zig std command](https://ziglang.org/download/0.12.0/release-notes.html#Installed-Standard-Library-Documentation) which is pretty neat! Therefore, I don't think it makes sense to support the Zig Standard Library beyond Zig 0.11.0.

See https://github.com/puqeko/zig2dash/issues/1#issuecomment-2525677687_

PR #5028, opened 2024-07-02
feat: add OpenUI5 docset in version 1.108.32

monavari-lebrecht:


docsets/OpenUI5/docset.json
```diff
@@ -1,10 +1,10 @@
 {
     "name": "OpenUI5",
-    "version": "1.96.7",
+    "version": "1.108.32",
     "archive": "UI5.tgz",
```
Kapeli, line 4: This file doesn't exist. You added `OpenUI5.tgz` instead I think.

Kapeli:
Thank you!

PR #5147, opened 2024-09-14
Add equinox 0.11.5

garymm:


garymm:
If there's an easy way to include all the required js and fonts in the docset (to avoid the WARNING in the test), that would be good to document. I basically just ran doc2dash

Kapeli:
> If there's an easy way to include all the required js and fonts in the docset (to avoid the WARNING in the test), that would be good to document. I basically just ran doc2dash

Making doc pages work 100% offline is quite hard, especially when it comes to big JS libraries like MathJax. I'm not aware of an easy way currently.

PR #5185, opened 2024-10-14
Dlang docset updated - 2.109

kassane:


Kapeli:
Thank you!

dejlek:
**Thank you Matheus!** 

I just hope one day Zeal author(s) add dedicated docset search field, so I do not have to prefix with `D:` all the time... :)

PR #5198, opened 2024-10-21
Update Godot4 to 4.3; Add specific version 4.2

scillidan:
The Godot4 docset is too big.

I upload it on [MEGAsync](https://mega.nz/folder/k1wVnIga#APeOgLgeWePzSNVPKTzxQQ). And keep the directory tree.


Kapeli:
Can you take over the existing Godot docset please? You can move the 3.x version to a specific version.

Also, is 4.2 required or could we skip it? Considering the large size of the docsets, keeping older versions might become problematic. I can accept it for now, but I might remove it later on, when space becomes an issue.

scillidan:
I originally planned to upload Godot3 later, but I wasn't sure how to submit two PRs at the same time. I moved the original Godot to Godot3, which is version 3.5, and 3.6 is the final version. But based on the level of citation it is in the community, I think 3.6 is not necessary. I also think it's better to keep only one stable version Godot4. At present, it's 4.3. What do I need to do? I can resubmit the PR.

Kapeli:
What I mean is add Godot 4 as an update to https://github.com/Kapeli/Dash-User-Contributions/tree/master/docsets/Godot, not in a separate folder of its own. Would this be ok?

scillidan:
Yes, it's ok. I will re-create the PR.

PR #5265, opened 2024-12-13
Adds latest Xojo Documentation

patricksalo:
Adds documentation for Xojo 2024r2, 2024r3 and 2024r4 and makes 2024r4 the new default.


Kapeli:
Thank you!

Kapeli:
I've removed 2024r3 and 2024r2, to save space. Unfortunately I can't keep around minor versions for large docsets.

patricksalo:
> I've removed 2024r3 and 2024r2, to save space. Unfortunately I can't keep around minor versions for large docsets. Sorry!

Okay makes sense and moving forward I'll ensure that there's just a single Xojo DocSet for each year. Your latest commit also removed 2024r1 which is great as this sets us up for this paradigm.

What is the replication time for this? Right now in Dash I can see 2024r4, 2024r3, 2024r2, 2024r1, & 2023r4 listed, of which 2024r4, 2024r1 & 2023r4 is downloadable. I believe we want to get to the point where only 2024r4 & 2023r4 are listed and also downloadable.

patricksalo:
Looks like just a replication delay as today I'm just seeing 2024r4 and 2023r4 in Dash. We're all set and thanks!

PR #5268, opened 2024-12-14
UnicodeCharacters: Misc. improvements

remko:
- Show character in search
- Make characters searchable by alias
- Make characters searchable by character
- Skip representation of non-displayable characters


Kapeli:
Thank you!

PR #5340, opened 2025-02-19
Bump boto3 to 1.36.23

jamesoff:
Binary file is at https://d24v0t7zqzj9pw.cloudfront.net/1.36.23/boto3.tgz for 90d.

Thanks!


jamesoff:
Think I broke something in my workflow; will redo this.

PR #5363, opened 2025-03-08
Add VHDL93 docset

Var7600:
Adding docset for VHDL93 - https://var7600.github.io/VHDL93-Docset/


Kapeli:
Hi. The `Info.plist` file is case-sensitive (you need to rename from `info.plist` to `Info.plist`).

Also, you need to add the following to the `Info.plist`: `<key>isDashDocset</key><true/>`

Thanks for the contribution!

Kapeli:
This failed validation. There are entries in `docSet.dsidx` which contain new lines in their names. This is not allowed.

Var7600:
Hey, Thanks you! for your help and guide. 
if there's anything else that i need to do let me know.

PR #5474, opened 2025-08-13
Update Ramda to 0.31.3

raszi:
Related to https://github.com/alexbepple/ramda-dash-docset/pull/26


docsets/Ramda/docset.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Ramda",
-  "version": "0.27.0",
+  "version": "0.31.3",
   "archive": "Ramda.tgz",
```
Kapeli, line 4: This file needs to be update as well

PR #5539, opened 2025-11-29
Add/Update Raycast docset to 1.0.0

ichoosetoaccept:
Adds/updates the Raycast docset.

Generated using https://github.com/ichoosetoaccept/raycast-docset


ichoosetoaccept:
Minor fix: Changed regex to use `\s+` instead of single space to handle attributes preceded by tabs, newlines, or multiple spaces (line-broken HTML).

PR #5606, opened 2026-04-14
GitHub_Actions.docset

toy:
I'm not sure if GitHub documentation LICENSE.txt is needed, as it is included in the docset anyway


Kapeli:
Thank you!

PR #5608, opened 2026-04-16
[codex] Add Modal docset

sheikheddy:
## Summary

- add a new `Modal` user-contributed Dash docset
- include the generated `Modal.tgz` archive, contribution icons, and `docset.json` metadata
- include `build_modal_docset.py` plus README instructions so the docset can be rebuilt from Modal docs and the GPU glossary

## Why

Modal does not currently have a user-contributed Dash docset, and the generated archive makes the platform docs usable offline in Dash.

## Validation

- verified the contribution icons are `16x16` and `32x32`
- verified `Modal.tgz` contains `Modal.docset/` with `Info.plist`, `docSet.dsidx`, and mirrored docs content
- ran `python3 -m py_compile docsets/Modal/build_modal_docset.py`


Kapeli:
Thank you!

PR #5611, opened 2026-04-18
Rubocop docsets

toy:
Not sure how to deal with with taking over maintenance of docsets, I tried using [scripts that were used before](https://github.com/milouse/dash-rubocop), but without success, so went with adapting [antora](https://docs.antora.org/) - the tool [used by rubocop developers](https://docs.rubocop.org/). Also generating docs for [all plugins](https://github.com/rubocop/docs.rubocop.org/blob/master/antora-playbook.yml) that are included on docs site. I also didn't keep the `specific_versions`
cc: @milouse


Kapeli:
Thank you!

PR #5653, opened 2026-07-02
[codex] Update PyTorch docset to 2.11.0

sheikheddy:
## Summary

- update the PyTorch user-contributed docset from 2.4.0 to 2.11.0
- add 2.11.0 as a selectable specific version while preserving the existing Yue Gao versions
- generate the archive from PyTorch's official 2.11 HTML documentation with `doc2dash`
- remove duplicated website navigation chrome from each page so the archive stays below GitHub's file-size limit and reads cleanly in Dash
- bundle KaTeX assets and referenced content images locally, and remove analytics/widgets that require network access

## Source

- documentation: https://github.com/pytorch/docs/tree/site/2.11
- PyTorch docset target: https://github.com/pytorch/pytorch/blob/v2.11.0/docs/Makefile

## Validation

- Kapeli `docsetcontrib --verify`: PASS for both archive paths
- SQLite integrity check: `ok`
- 14,632 search entries; no empty names, missing files, or missing fragments
- verified representative entries including `torch.compile`, `torch.Tensor`, `torch.nn.Module`, and `torch.accelerator`
- verified 16x16 and 32x32 icons
- verified no active external scripts, stylesheets, fonts, images, or media remain
- both archives are byte-identical, 59,211,445 bytes, SHA-256 `77576569ff1398ef8ae47015a26d1333d31ff4b8ceab4bd5e9cfb30b1e58d028`

### DIFF
--- a/docsets/PyTorch/README.md
+++ b/docsets/PyTorch/README.md
@@ -11,5 +11,6 @@ PyTorch Docset
 
 #### Docset Author:
 
+- [sheikheddy](https://github.com/sheikheddy)
 - [Yue Gao](https://github.com/hologerry)
 - [Aziz Alto](https://github.com/iamaziz)
--- a/docsets/PyTorch/docset.json
+++ b/docsets/PyTorch/docset.json
@@ -1,10 +1,10 @@
 {
     "name": "PyTorch",
-    "version": "2.4.0",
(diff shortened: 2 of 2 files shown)


sheikheddy:
@Kapeli

PR #5696, opened 2026-09-02
Add ARM intrinsics docset

JoeyTeng:
The source is from the developing `main` branch rather that the `latest-release`; I can change that to be based on `latest-release` up on request.

The icon is generated by AI to prevent licensing cumbersome.

If it's OK I'd like to update the [Intel Intrinsics docset](https://github.com/Kapeli/Dash-User-Contributions/tree/master/docsets/Intel_Intrinsics) as well.

Also if possible I'd like to have a free license; I have an existing life-time license but that's for an old version...


Kapeli:
Please don't include the source of the docset, just the packaged docset.

JoeyTeng:
The PR now contains:

- Packaged docset
- Reproducible generator
- Licensing and attribution

@Kapeli is this OK or will you prefer to remove more files from this PR?

